PR #9, opened 2016-03-24
client message builder implementation. Client message fragments merge…

asimarslan:
… into actuall messages

### DIFF
--- a/hazelcast/protocol/client_message.py
+++ b/hazelcast/protocol/client_message.py
@@ -23,6 +23,7 @@ Any request parameter, response or event data will be carried in the payload.
 
 """
 import binascii
+import logging
 import struct
 
 from hazelcast.serialization.data import *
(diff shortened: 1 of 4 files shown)


hazelcast/protocol/client_message.py
```diff
@@ -223,6 +233,22 @@ def __str__(self):
...
+        self._message_callback = message_callback
+
+    def on_message(self, client_message):
+        if client_message.is_flag_set(BEGIN_END_FLAG):
+            # handle message
+            self._message_callback(client_message)
+        elif client_message.is_flag_set(BEGIN_FLAG):
+            self._incomplete_messages[client_message.get_correlation_id()] = client_message
+        else:
+            message = self._incomplete_messages[client_message.get_correlation_id()]
```
cangencer: might make sense to add a catch here with a warning, so that it doesn't break reactor thread if it receives a message  which it didn't receive the first part.

cangencer:
:+1: , except one minor point.

PR #12, opened 2016-05-17
Member equals comparison should include both UUID and address

### LINKED ISSUE #10: Members list of cluster not refreshed after reconnect
Given the following scenario: 
- A cluster made up of a single Java Hazelcast member.
- A single Hazelcast Python client that connects to the cluster through this Java member.

If the Hazelcast member gets disconnected or crashes, the Python client attempts to reconnect a few times (refer to the function ClusterService._reconnect at cluster.py) and if the Hazelcast member goes back online soon enough, the Python client nicely reconnects. 

The problem is that quite possibly, the Hazelcast member ID changes after being restarted and for some reason this change does not get reflected in the list of members kept by the Python client (specifically,  ClusterService.members). Thus, if we get a message say, on a queue, and we attempt to get the member that delivered the message through:

hz_python_client.cluster.get_member_by_uuid(cur_msg.issuingMemberId)

We will get None, even though cur_msg.issuingMemberId is part of the cluster.

I'm not really sure why, but the issue gets fixed by clearing the list of members before attempting to reconnect. We did so by modifying the function ClusterService._reconnect like this:

    def _reconnect(self):
        try:
            self.logger.warn("Connection closed to owner node. Trying to reconnect.")
            del self.members[:]
            self._connect_to_cluster()

This slight modification causes the event ClusterService_handle_member to be triggered immediately after reconnecting, which in turn properly updates the members list. Certainly, this looks more as a hack than as a nice solution. Could you please check it out?

Thank you!

cangencer:
Fixes #10

### DIFF
--- a/hazelcast/core.py
+++ b/hazelcast/core.py
@@ -19,7 +19,7 @@ class Member(object):
             .format(self.address.host, self.address.port, self.uuid, self.is_lite_member, self.attributes)
 
     def __eq__(self, other):
-        return isinstance(other, self.__class__) and self.address == other.address
+        return isinstance(other, self.__class__) and self.address == other.address and self.uuid == other.uuid
 
 
 class Address(object):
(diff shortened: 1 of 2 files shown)


cangencer:
verify

asimarslan:
👍

PR #16, opened 2016-08-11
Portable serialization version support (Morphing portable reader impl.)

asimarslan:


hazelcast/serialization/portable/classdef.py
```diff
@@ -104,7 +102,7 @@ def __str__(self):
         return "fid:{}, cid:{}, v:{}, fields:{}".format(self.factory_id, self.class_id, self.version, self.field_defs)
 
     def __repr__(self):
-        return repr(self)
+        return self.__str__()
```
cangencer: I think this is not needed, we should define either __repr__ or __str__.

__str__ will default to __repr__ if __repr__ is missing, so we should just define __repr__ and not define __str__. Also the python convention for __repr__ is  TypeName(field=value,field2=value2...), i.e.: https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/proxy/base.py#L140

cangencer:
👍  except some comments regarding str and repr usage.

PR #17, opened 2016-09-01
Api Documentation for map.py

oguzdemir:
Map.py is documented for Python Sphinx.


hazelcast/proxy/map.py
```diff
@@ -21,9 +21,30 @@
...
+
+    This class does not allow None to be used as a key or value.
     """
     def add_entry_listener(self, include_value=False, key=None, predicate=None, added=None, removed=None, updated=None,
                            evicted=None, evict_all=None, clear_all=None, merged=None, expired=None):
+        """
+        Adds an continuous entry listener for this map. Listener will get notified for map events filtered with given
+        parameters.
+
+        :param include_value:
```
asimarslan: received events should include an old value or not

PR #27, opened 2016-09-29
ApiDoc for NearCache, Partition and Transaction

oguzdemir:
ApiDoc added to near_cache.py, partition.py and transaction.py with Python Sphinx format.

### DIFF
--- a/hazelcast/partition.py
+++ b/hazelcast/partition.py
@@ -7,6 +7,9 @@ PARTITION_UPDATE_INTERVAL = 10
 
 
 class PartitionService(object):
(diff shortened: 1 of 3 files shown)


hazelcast/partition.py
```diff
@@ -7,6 +7,9 @@
 
 
 class PartitionService(object):
+    """
+    A SPI service for accessing partition related information.
```
Serdaro: An SPI service...

oguzdemir: Done.

PR #54, opened 2017-02-22
check exception type in asyncore.dispatcher handle_error

sancar:
fixes https://github.com/hazelcast/hazelcast-python-client/issues/58


sancar:
verify

sancar:
verify

PR #56, opened 2017-02-22
authenticate manager should send uuid's if available.

sancar:
authenticate manager should send uuid's if available. 
Initialise lastRead time of connection in connection constructor so that it will not be immediately closed down by heartbeat mechanism

fixes https://github.com/hazelcast/hazelcast-python-client/issues/59
fixes https://github.com/hazelcast/hazelcast-python-client/issues/60


sancar:
verify

PR #65, opened 2017-04-18
Client hangs on member shutdown

### LINKED ISSUE #64: Client hangs on member shutdown
Partition count refresher does not respect connection shutdown.

asimarslan:
fixes #64


hazelcast/reactor.py
```diff
@@ -63,6 +64,9 @@ def _check_timers(self):
                 return
 
     def add_timer_absolute(self, timeout, callback):
+        # if not self._is_live:
```
sancar: Why this part is commented out ?

PR #85, opened 2018-05-28
fixing high cpu usage on asycore.loop

### LINKED ISSUE #83: High CPU usage caused by the Hazelcast client
Ever since we added the HZ client to one of our components, CPU usage of that component increased significantly. We could track it down to the `select` syscall, which is used by [`asyncore.loop` called by AsyncoreReactor](https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/reactor.py#L37).

CPU usage could be reduced by changing that call to `loop(count=1, timeout=0.1)`.

Our application itself makes heavy use of asyncore, so running two threads in one process, each with its own asyncore main loop in it, doesn't really make much sense. We'd appreciate being able to have the HZ client use our main loop, instead of living on its own thread. Would that be feasible?

sancar:
fixes #83

### DIFF
--- a/hazelcast/reactor.py
+++ b/hazelcast/reactor.py
@@ -34,7 +34,7 @@ class AsyncoreReactor(object):
         Future._threading_locals.is_reactor_thread = True
         while self._is_live:
             try:
-                asyncore.loop(count=10, timeout=0.01, map=self._map)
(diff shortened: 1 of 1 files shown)


sancar:
verify

PR #95, opened 2018-07-19
updated documentation, added soak test and renamed await for python3 support changes

mdumandag:
- Updated documentation to specify that client is now also supporting Python3.4+

- Added soak test for map

- Renamed await method of CountDownLatch with await_latch since await is a reserved keyword for Python3.7


asimarslan:
verify

PR #96, opened 2018-07-24
fix pop behaviour of connection manager

mdumandag:
When client authenticates with the same address more than once,  [connections](https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/connection.py#L36) and [_pending_connections](https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/connection.py#L37) dictionaries of the connection manager are updated with the latests address-future pair as it should be. However, when on_auth or _connection_closed methods are called for each connection, these methods try to pop the same address from these dictionaries more than once which results in a KeyError.

We should try to pop the address but when KeyError happens we should simply pass since that address is already removed from these dictionaries by the previous method calls.


mdumandag:
verify

PR #104, opened 2018-08-13
reactor Timer ordering

mmsolutions:
Hi,

Mostly my script is hanging when working on blocking map immediately after Hazlecast client created.
Below stack trace:

ERROR:Reactor:Error in Reactor Thread
  File "C:\Tools\Python36\lib\site-packages\hazelcast\reactor.py", line 38, in _loop
    self._check_timers()
  File "C:\Tools\Python36\lib\site-packages\hazelcast\reactor.py", line 60, in _check_timers
    self._timers.get_nowait()
  File "C:\Tools\Python36\lib\queue.py", line 192, in get_nowait
    return self.get(block=False)
  File "C:\Tools\Python36\lib\queue.py", line 174, in get
    item = self._get()
  File "C:\Tools\Python36\lib\queue.py", line 230, in _get
    return heappop(self.queue)
TypeError: '<' not supported between instances of 'Timer' and 'Timer'

When using pause time.sleep(1) before map consuming script works perfectly.

I think that the reason is reactor initialization, that stores Timers in queue.PriorityQueue() but Timer class is not comparable.
So I have added total ordering from functools and it works without problems.

Regards,
MM

### DIFF
--- a/hazelcast/reactor.py
+++ b/hazelcast/reactor.py
@@ -6,8 +6,10 @@ import socket
 import sys
 import threading
 import time
-from hazelcast.six.moves import queue
 from collections import deque
+from functools import total_ordering
+
+from hazelcast.six.moves import queue
 
 from hazelcast.connection import Connection, BUFFER_SIZE
 from hazelcast.exception import HazelcastError
@@ -191,6 +193,7 @@ class AsyncoreConnection(Connection, asyncore.dispatcher):
             self._connection_closed_callback(self, cause)
 
 
+@total_ordering
 class Timer(object):
     canceled = False
 
@@ -198,6 +201,12 @@ class Timer(object):
         self.end = end
         self.timer_ended_cb = timer_ended_cb
         self.timer_canceled_cb = timer_canceled_cb
+
+    def __eq__(self, other):
+        return self.end == other.end
+
+    def __ne__(self, other):
+        return not (self == other)
 
     def __lt__(self, other):
         return self.end < other.end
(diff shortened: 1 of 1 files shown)


mdumandag:
Hi @mmsolutions 
Your reasoning seems true to me. On Windows time.time() has a bad resolution compared to UNIX systems. So, probably time.time() returns the same number in consecutive calls and priority queue comparison falls back to Timer's itself which is not comparable.

I checked the implementation of priority queue.  It only compares elements by <. So, it is enough to just implement \_\_lt\_\_ for this case. 

Can you update your PR to add just the \_\_lt\_\_ method? After that, I can start the tests and merge your PR.

Thanks for your contribution

mmsolutions:
Of course, done.

mdumandag:
verify

mdumandag:
verify

sancar:
verify

cangencer:
I think it would be better to implement __gt__ and __eq__ as well, as you shouldn't rely on the fact that priority queue only checks for less than.

mmsolutions:
Yep, that was the first commit to make Timer comparable in a full manner.
@mdumandag what do you think? Do you want to revert second one?

mdumandag:
Okay, let's make it fully comparable. @mmsolutions Can you revert the second commit ?

mmsolutions:
Of course @mdumandag, done.

mdumandag:
verify

sancar:
verify

PR #110, opened 2018-08-31
Fix lock timeout behaviour of when key owner is gone

### LINKED ISSUE #107: Lock operations get timeout exception when member dies
When a client waiting for a lock to be released more than invocation timeout seconds, and the member dies, clients get operation timeout exception rather than it is redirected to another member.

issue from java client https://github.com/hazelcast/hazelcast/issues/13551
fix to java client           https://github.com/hazelcast/hazelcast/pull/13552

mdumandag:
When a client waiting on lock to get it, if the member that it is waiting on shuts down/terminated , client retries it on new owner of the lock.

If a client waits for longer than client.invocation.timeout.seconds, when exception came from server, it is wrapped inside OperationTimeoutException and thrown to user.

This pr changes this behaviour so that waiting lock operations (map.lock/try_lock lock.lock/try_lock)  will not throw exception but retry the operation.

fixes #107


tests/proxy/lock_test.py
```diff
@@ -80,4 +80,41 @@ def test_force_unlock(self):
...
+        self.start_new_thread(lock_thread_func)
+
+        time.sleep(2 * invocation_timeout_seconds)
+        key_owner.shutdown()
+
+        partition_id = self.client.partition_service.get_partition_id(key)
+        while not (self.client.partition_service.get_partition_owner(partition_id) == self.member.address):
+            time.sleep(0.1)
+
+        self.assertTrue(server_lock.is_locked())
```
SaitTalhaNisanci: If this assertion fails, will this code continue or abort ? If it aborts since the revert part wont be done other tests could fail as well.

mdumandag: good point. fixed it

PR #115, opened 2018-09-19
implement client properties

### LINKED ISSUE #114: implement client properties
Right now, configuration properties are set with constant strings that are defined on top of the module they are being used. This causes properties to scatter all over the code and makes it hard for user to import them one by one.

There should be a common class that holds all the available configuration properties to deal with that problem.

mdumandag:
this pr implements client properties so that we can access all the available properties on one common class

fixes #114

### DIFF
--- a/hazelcast/config.py
+++ b/hazelcast/config.py
@@ -2,9 +2,10 @@
 Hazelcast Client Configuration module contains configuration classes and various constants required to create a ClientConfig.
 
 """
+import os
 
 from hazelcast.serialization.api import StreamSerializer
-from hazelcast.util import validate_type, validate_serializer, enum
+from hazelcast.util import validate_type, validate_serializer, enum, TimeUnit
 
 DEFAULT_GROUP_NAME = "dev"
 """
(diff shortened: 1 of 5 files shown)


hazelcast/config.py
```diff
@@ -439,3 +447,81 @@ def eviction_sampling_pool_size(self, eviction_sampling_pool_size):
...
+    def get_seconds(self, property):
+        """
+        Gets the value of the given property in seconds. If the value of the given property is not a number,
+        throws TypeError.
+        :param property: (:class:`~hazelcast.config.ClientProperty`), Property to get seconds from
+        :return: (float), Value of the given property in seconds
+        """
+        return TimeUnit.to_seconds(self.get(property), property.time_unit)
+
+    def get_seconds_positive(self, property):
```
SaitTalhaNisanci: Could a name like get_seconds_positive_or_default be better ? It seems long but it describes what the function does better.

hazelcast/util.py
```diff
@@ -219,3 +219,27 @@ def get_portable_version(portable, default_version):
...
         version = default_version
     return version
+
+
+class TimeUnit(object):
+    """
+    Represents the time durations at given units in seconds.
+    """
+    NANOSECOND = 1e-9
+    MICROSECOND = 1e-6
```
SaitTalhaNisanci: I think it is better to express the units with respect to each other. For example instead of `MICROSECOND = 1e-6`, we can use `MICROSECOND = NANOSECOND *1000` and so on.

PR #120, opened 2018-10-18
update documentation according to sphinx formatting

mdumandag:
Updated doc strings to successfully build API docs using Sphinx


mdumandag:
verify

PR #128, opened 2018-11-08
change debug level to info in lifecycle service

SaitTalhaNisanci:
Firing a lifecycle event should have `info` level instead of `debug`.

### DIFF
--- a/hazelcast/lifecycle.py
+++ b/hazelcast/lifecycle.py
@@ -58,7 +58,7 @@ class LifecycleService(object):
             self.is_live = False
 
         self.state = new_state
-        self.logger.debug("New Lifecycle state is %s", new_state)
+        self.logger.info("New Lifecycle state is %s", new_state)
         for listener in list(self._listeners.values()):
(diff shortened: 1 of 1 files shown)


hazelcast/lifecycle.py
```diff
@@ -58,7 +58,7 @@ def fire_lifecycle_event(self, new_state):
             self.is_live = False
 
         self.state = new_state
-        self.logger.debug("New Lifecycle state is %s", new_state)
+        self.logger.info("New Lifecycle state is %s", new_state)
```
mdumandag: Can you also update the following outputs according to this new change?

- https://github.com/hazelcast/hazelcast-python-client#15-basic-usage
- https://github.com/hazelcast/hazelcast-python-client#791-logging-configuration
- https://github.com/hazelcast/hazelcast-python-client#setting-custom-handlers

Holmistr:
verify

mdumandag:
fixed in #146

PR #130, opened 2018-11-09
Add a test in which client trust default ca stores

mdumandag:
Added a test where SSL is enabled but no cafile is set.
In this case, client will try to load default ca certificates
and if the member's certificate is signed by a trusted ca,
connection will be established. For this test, a certificate
from Let's Encrypt will be used.


run-tests.ps1
```diff
@@ -1,20 +1,73 @@
...
+if(Test-Path env:HAZELCAST_ENTERPRISE_KEY){
+	if(Test-Path "hazelcast-enterprise-${hazelcastEnterpriseVersion}.jar") {
+		Write-Host "hazelcast-enterprise-${hazelcastEnterpriseVersion}.jar already exist, not downloading from maven."
+	} else {
+		Write-Host "Downloading hazelcast-enterprise-${hazelcastEnterpriseVersion}.jar ..."
+		& "mvn" -q "dependency:get" "-DrepoUrl=${enterpriseRepo}" "-Dartifact=com.hazelcast:hazelcast-enterprise:${hazelcastEnterpriseVersion}" "-Ddest=hazelcast-enterprise-${hazelcastEnterpriseVersion}.jar"
+	}
+
+	if(Test-Path "hazelcast-enterprise-${hazelcastEnterpriseTestVersion}-tests.jar") {
+		Write-Host "hazelcast-enterprise-${hazelcastEnterpriseVersion}-tests.jar already exist, not downloading from maven."
```
asimarslan: should be hazelcastEnterpriseTestVersion I guess

PR #135, opened 2018-11-23
Use snapshot version of the enterprise tests

mdumandag:
Use snapshot enterprise tests JAR to access the renewed
Let's Encrypt certificate


mdumandag:
verify

PR #136, opened 2018-12-10
implement client statistics

mdumandag:
this pr implements the client statistics feature.

client and near cache related statistics are collected by default. To get the OS and runtime statistics, an extra dependency is used. (psutil).


hazelcast/util.py
```diff
@@ -66,6 +66,14 @@ def current_time():
     return time.time()
 
 
+def current_time_in_millis():
+    """
+    Returns the current time of the system in millis.
+    :return: (int), current time of the system in millis.
+    """
+    return int(time.time() * 1000)
```
ihsandemir: can you use a constant specifier instead of number 1000 here so that it is more understandable? E.g. MILLISECONDS_IN_A_SECOND

hazelcast/util.py
```diff
@@ -87,6 +95,18 @@ def to_millis(seconds):
...
 
+def to_nanos(seconds):
+    """
+    Converts the time parameter in seconds to nanoseconds. If the given time is negative, returns the original value.
+
+    :param seconds: (Number), the given time in seconds.
+    :return: (int), result of the conversation in nanoseconds.
+    """
+    if seconds >= 0:
+        return int(seconds * 1e9)
```
ihsandemir: similarly instead of 1e9, better to use a constant.

tests/client_message_test.py
```diff
@@ -171,7 +171,7 @@ def test_set_flag_listener(self):
 
 class ClientMessageBuilderTest(unittest.TestCase):
     def test_message_accumulate(self):
-        message = client_authentication_codec.encode_request("user", "pass", "uuid", "owner-uuid", True, "PYH", 1)
+        message = client_authentication_codec.encode_request("user", "pass", "uuid", "owner-uuid", True, "PYH", 1, "3.10")
```
ihsandemir: use CLIENT_VERSION instead of 3.10? it will fail in the next release.

tests/statistics_test.py
```diff
@@ -0,0 +1,239 @@
...
+        config.set_property(ClientProperties.STATISTICS_ENABLED.name, True)
+        config.set_property(ClientProperties.STATISTICS_PERIOD_SECONDS.name, self.STATS_PERIOD)
+        client = HazelcastClient(config)
+
+        time.sleep(2 * self.STATS_PERIOD)
+        response1 = self._get_client_stats_from_server()
+        self.assertTrue(response1.success)
+        self.assertIsNotNone(response1.result)
+
+        time.sleep(2 * self.STATS_PERIOD)
```
ihsandemir: time dependent test may fail, better to use `eventually` logic (e.g. `assertTrueEventually`).

tests/statistics_test.py
```diff
@@ -0,0 +1,239 @@
...
+
+        response = self._get_client_stats_from_server()
+        self.assertTrue(response.success)
+        self.assertIsNotNone(response.result)
+
+        result = response.result.decode("utf-8")
+        local_address = self._get_local_address(client)
+
+        # Check near cache ve client statistics
```
ihsandemir: `ve`? did you mean `and`?

tests/statistics_test.py
```diff
@@ -0,0 +1,239 @@
...
+        self.assertEqual(1, result.count("clientName=" + client.name))
+        self.assertEqual(1, result.count("lastStatisticsCollectionTime="))
+        self.assertEqual(1, result.count("enterprise=false"))
+        self.assertEqual(1, result.count("clientType=" + CLIENT_TYPE))
+        self.assertEqual(1, result.count("clientVersion=" + CLIENT_VERSION))
+        self.assertEqual(1, result.count("clusterConnectionTimestamp="))
+        self.assertEqual(1, result.count("clientAddress=" + local_address))
+        self.assertEqual(1, result.count("nc." + map_name + ".creationTime"))
+        self.assertEqual(1, result.count("nc." + map_name + ".evictions"))
```
ihsandemir: you can also validate the actual values for near cache stats, can you add them?

hazelcast/near_cache.py
```diff
@@ -193,9 +196,19 @@ def _is_eviction_required(self):
     def get_statistics(self):
         """
         Returns the statistics of the NearCache.
-        :return: (Number, Number), evicted entry count and expired entry count.
+        :return: (Dict), Dictionary that stores statistics related to this near cache.
         """
-        return self._evicted_count, self._expired_count
+        stats = {
+            "creation_time": self._creation_time,
```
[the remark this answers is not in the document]
mdumandag: owned entry memory is cost is size in this case.I will change its name. Persistance related statistics are not collected since near cache preloader is not implemented yet. 
I can't see the invalidation related statistics in the design document or the java doc string. Are they added later ? https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngine.java#L83

hazelcast/protocol/codec/client_authentication_codec.py
```diff
@@ -41,13 +44,14 @@ def encode_request(username, password, uuid, owner_uuid, is_owner_connection, cl
...
     client_message.append_byte(serialization_version)
+    client_message.append_str(client_hazelcast_version)
     client_message.update_frame_length()
     return client_message
 
 
 def decode_response(client_message, to_object=None):
     """ Decode response from client message"""
-    parameters = dict(status=None, address=None, uuid=None, owner_uuid=None, serialization_version=None)
+    parameters = dict(status=None, address=None, uuid=None, owner_uuid=None, serialization_version=None, server_hazelcast_version=None, client_unregistered_members=None)
```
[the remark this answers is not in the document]
mdumandag: My local copy of the client protocol may be outdated. I will fix this

hazelcast/protocol/codec/map_key_set_codec.py
```diff
@@ -6,7 +6,7 @@
 
 REQUEST_TYPE = MAP_KEYSET
 RESPONSE_TYPE = 106
-RETRYABLE = False
+RETRYABLE = True
```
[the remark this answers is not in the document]
mdumandag: Okay, I will separate them

hazelcast/proxy/map.py
```diff
@@ -898,12 +898,9 @@ class MapFeatNearCache(Map):
...
     def __init__(self, client, service_name, name):
         super(MapFeatNearCache, self).__init__(client, service_name, name)
-        near_cache_config = client.config.near_cache_configs.get(name, None)
-        if near_cache_config is None:
-            raise ValueError("NearCache config cannot be None here!")
         self._invalidation_listener_id = None
-        self._near_cache = create_near_cache(client.serialization_service, near_cache_config)
-        if near_cache_config.invalidate_on_change:
+        self._near_cache = client.near_cache_manager.get_or_create_near_cache(name)
+        if self._near_cache.invalidate_on_change:
```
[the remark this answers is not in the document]
mdumandag: Within the NearCacheManager, I get the config from the client.config.near_cache_configs using the name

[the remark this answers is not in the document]
mdumandag: I used the near cache config within the near cache manager.  So, I removed the line which we get the near cache config and used near cache directly instead of the config. It is initialized with the given config so nothing changes

hazelcast/reactor.py
```diff
@@ -182,10 +182,12 @@ def __init__(self, map, address, connect_timeout, socket_options, connection_clo
         self._write_queue.append(b"CB2")
 
     def handle_connect(self):
+        self.start_time = time.time()
         self.logger.debug("Connected to %s", self._address)
 
     def handle_read(self):
         self._read_buffer += self.recv(BUFFER_SIZE)
+        self.last_read = time.time()
```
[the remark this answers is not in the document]
mdumandag: I removed it from the connection.py and get the last read time here since recv is done here. There is no duplicate

hazelcast/statistics.py
```diff
@@ -0,0 +1,251 @@
...
+                                "{} instead.".format(ClientProperties.STATISTICS_PERIOD_SECONDS.name,
+                                                     period,
+                                                     default_period))
+            period = default_period
+
+        def _statistics_task():
+            self._send_statistics()
+            self._statistics_timer = self._client.reactor.add_timer(period, _statistics_task)
+
+        self._statistics_timer = self._client.reactor.add_timer(period, _statistics_task)
```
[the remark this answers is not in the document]
mdumandag: It is a recursive call. So, when the task sends the stats to the member, It will add another timer to make it periodic.

hazelcast/statistics.py
```diff
@@ -0,0 +1,251 @@
...
+        if PSUTIL_ENABLED:
+            try:
+                # Availability: All platforms. Try to be sure that it is working.
+                if "phy_mem" not in self._failed_gauges:
+                    memory_info = psutil.virtual_memory()
+                    psutil_stats["os.totalPhysicalMemorySize"] = memory_info.total
+                    psutil_stats["os.freePhysicalMemorySize"] = memory_info.available
+            except Exception as ex:
+                self.logger.warning(self._EXCEPTION_STR.format("physical memory size", ex))
+                self._failed_gauges.add("phy_mem")
```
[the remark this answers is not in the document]
mdumandag: See https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L478 for example. When available memory cannot be determined, psutil will not return it so we can get AttributeError here.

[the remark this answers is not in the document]
mdumandag: See https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L402 here for example. When the `/proc/meminfo` doesn't contain the fields, we can get a KeyError. I think it is best to call them within the try catch block to avoid such edge cases. try catch block adds a very little overhead.

hazelcast/statistics.py
```diff
@@ -0,0 +1,251 @@
...
+            process = psutil.Process()
+            with process.oneshot():
+                # With oneshot, process related information could be gathered
+                # faster due to caching.
+
+                # Availability: All platforms. Try to be sure that it is working.
+                try:
+                    if "p_mem_info" not in self._failed_gauges:
+                        p_memory_info = process.memory_info()
+                        psutil_stats["os.committedVirtualMemorySize"] = p_memory_info.vms
```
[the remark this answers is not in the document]
mdumandag: See the design document. For example, this stat is described as `os.committedVirtualMemorySize: The amount of virtual memory that is guaranteed to be available to the running process in bytes, or -1 if this operation is not supported.` . So, it is not really os related.

hazelcast/statistics.py
```diff
@@ -0,0 +1,251 @@
...
+                        psutil_stats["runtime.usedMemory"] = p_memory_info.rss
+                except Exception as ex:
+                    self.logger.warning(self._EXCEPTION_STR.format("memory info related to the process", ex))
+                    self._failed_gauges.add("p_mem_info")
+
+                try:
+                    # Availability: UNIX
+                    if "p_fdc" not in self._failed_gauges:
+                        psutil_stats["os.openFileDescriptorCount"] = process.num_fds()
```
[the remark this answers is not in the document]
mdumandag: Again, this is not the open fdc of the whole system. It is related to the process

hazelcast/statistics.py
```diff
@@ -0,0 +1,251 @@
...
+
+                try:
+                    # Availability: All platforms. Try to be sure that it is working.
+                    if "p_uptime" not in self._failed_gauges:
+                        psutil_stats["runtime.uptime"] = to_millis(current_time() - process.create_time())
+                except Exception as ex:
+                    self.logger.warning(self._EXCEPTION_STR.format("process uptime", ex))
+                    self._failed_gauges.add("p_uptime")
+
+        return psutil_stats
```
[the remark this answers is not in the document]
mdumandag: These are documented in the PRD. See https://hazelcast.atlassian.net/wiki/spaces/PM/pages/236650623/Python+Client+Statistics

hazelcast/statistics.py
```diff
@@ -0,0 +1,251 @@
...
+        self._add_stat(stats, "enterprise", "false")
+        self._add_stat(stats, "clientType", CLIENT_TYPE)
+        self._add_stat(stats, "clientVersion", CLIENT_VERSION)
+        self._add_stat(stats, "clusterConnectionTimestamp", to_millis(owner_connection.start_time))
+
+        local_host, local_ip = owner_connection.socket.getsockname()
+        local_address = str(local_host) + ":" + str(local_ip)
+        self._add_stat(stats, "clientAddress", local_address)
+        self._add_stat(stats, "clientName", self._client.name)
+        self._add_stat(stats, "credentials.principal", self._client.config.group_config.name)
```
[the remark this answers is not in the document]
mdumandag: `UsernamePasswordCredentials ` is  not implemented in this client yet. So, I think this is the closest thing to that. Should I send this like that or remove it ?

tests/statistics_test.py
mdumandag: I will change it

tests/statistics_test.py
mdumandag: I am testing it in the `test_near_cache_stats `. Should I do it there too ?

tests/statistics_test.py
mdumandag: Okay, I will change it

tests/client_message_test.py
mdumandag: It is just a test to message builder. None of these parameters are validated. How could it fail ?

hazelcast/protocol/codec/client_authentication_codec.py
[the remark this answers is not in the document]
mdumandag: I checked it again using the `Hazelcast Client Protocol 1.8.0-2` but `client_unregistered_members` is still there. Other clients also have this field. See https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/proto/client_authentication.go#L69 https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/codec/ClientAuthenticationCodec.ts#L115 . Are you sure that there is no such field ?

hazelcast/statistics.py
```diff
@@ -0,0 +1,251 @@
...
+        os_and_runtime_stats = self._get_os_and_runtime_stats()
+        for stat in os_and_runtime_stats:
+            self._add_stat(stats, stat, os_and_runtime_stats[stat])
+
+    def _get_os_and_runtime_stats(self):
+        psutil_stats = {}
+        if PSUTIL_ENABLED:
+            try:
+                # Availability: All platforms. Try to be sure that it is working.
+                if "phy_mem" not in self._failed_gauges:
```
SaitTalhaNisanci: instead of writing the comment above, you can have a function such as:
```
def _can_collect_stat(name):
     return name not in self._failed_gauges
```
and here after removing the comment you can just have
 ```
 if self._can_collect_stat("phy_mem"):
```

and collecting the stat for each of this should be a separate function in my opinion. I believe the readability would be better if the layout of this function is as follows:

```
if self._can_collect_stat(statName1):
   self._collect_stat_name1()
if self._can_collect_stat(statName2):
   self._collect_stat_name2()
..
..
..
```

where `_collect_stat_name1()` will have the try catch block for the relevant stat.

**Comment by mdumandag** [comment]
verify

File: hazelcast/statistics.py [hunk 19]
```diff
@@ -0,0 +1,283 @@
...
+    def _add_runtime_and_os_stats(self, stats):
+        os_and_runtime_stats = self._get_os_and_runtime_stats()
+        for stat_name, stat_value in six.iteritems(os_and_runtime_stats):
+            self._add_stat(stats, stat_name, stat_value)
+
+    def _get_os_and_runtime_stats(self):
+        psutil_stats = {}
+        if PSUTIL_ENABLED:
+
+            # Availability: All platforms. Try to be sure that it is working.
```
SaitTalhaNisanci: I think we dont need this comment because the function name `_can_collect_stat` already includes the comment, if you think it doesnt you can rename it.  And realize that most of these comments are copy paste, so if the behavior of `_can_collect_stat` changes, it is very possible that some of these comments wont be updated, however if you choose the function name explanatory, which I already think it is, then changing function name would be enough.

hazelcast/util.py
```diff
@@ -83,7 +94,19 @@ def to_millis(seconds):
     :return: (int), result of the conversation in milliseconds.
     """
     if seconds >= 0:
-        return int(seconds * 1000)
+        return int(seconds * MILLISECONDS_IN_SECONDS)
+    return seconds
+
+
+def to_nanos(seconds):
```
SaitTalhaNisanci: Here the parameter is always in seconds right? So in that case why not name the function as `seconds_to_nanos(seconds)`. Also this function does not do what it says. As a caller when I call this function I expect to have nano seconds as the return value however if the given seconds is negative then it doesnt return nanoseconds. This is confusing. Imagine someone calling this in the future without realizing that? It could result in bugs at best. If this is really the behavior you want to get then name the function like that. But I think this behavior is weird. Can you explain why are we returning the original negative value in the second case?

tests/statistics_test.py
```diff
@@ -0,0 +1,243 @@
...
+        self.assertNotEqual(response1, response2)
+        client.shutdown()
+
+    def test_statistics_enabled_with_negative_period(self):
+        config = ClientConfig()
+        config.set_property(ClientProperties.STATISTICS_ENABLED.name, True)
+        config.set_property(ClientProperties.STATISTICS_PERIOD_SECONDS.name, -1 * self.STATS_PERIOD)
+        client = HazelcastClient(config)
+
+        time.sleep(2 * self.DEFAULT_STATS_PERIOD)
```
SaitTalhaNisanci: is it possible to use assert eventually here instead of time sensitive test?

tests/statistics_test.py
```diff
@@ -0,0 +1,243 @@
...
+        self.assertEqual(1, result.count("nc." + map_name + ".hits=1"))
+        self.assertEqual(1, result.count("nc." + map_name + ".misses=2"))
+        self.assertEqual(1, result.count("nc." + map_name + ".ownedEntryCount=1"))
+        self.assertEqual(1, result.count("nc." + map_name + ".expirations=0"))
+        self.assertEqual(1, result.count("nc." + map_name + ".invalidations=1"))
+        self.assertEqual(1, result.count("nc." + map_name + ".invalidationRequests=2"))
+
+        client.shutdown()
+
+    def _get_client_stats_from_server(self):
```
SaitTalhaNisanci: I know this is same for other clients too, but this can result in bugs in the future. If there are some leftover clients in the tests then getting the first client might not be correct. Nodejs client has updated this:

```
    var clientUuid = client.getClusterService().uuid;
    var script =
        'clients=instance_0.getClientService().getConnectedClients().toArray()\n' +
        'for(i=0;i<clients.length;i++) {\n' +
        '   if (clients[i].getUuid().equals("' + clientUuid + '")) {\n' +
        '       result=clients[i].getClientStatistics();\n' +
        '       break;' +
        '   }\n' +
        '}\n';
```

Doing this is a safe way instead of assuming that there wont be any leftover clients.

mdumandag: I will change this

tests/statistics_test.py
mdumandag: Okay, I will increase the assertion timeout to 30 seconds from 2 * stats period

hazelcast/statistics.py
mdumandag: I will remove the comments

hazelcast/util.py
mdumandag: I tried to mimic the behaviour of the old `to_millis` method without thinking too much about it. I think it's reasoning is that when you pass a negative time value it indicates the infinite or the default value in the context of Hazelcast time parameters (like TTL) so it is unnecessary to the multiplication here. But you are right this does not make sense. I will change them. For the naming, I think `to_millis` or `to_nanos` is just enough because time (and datetime) module returns everything as seconds. So, when you are doing a conversion to millis or nanos, it is almost always from seconds.

mdumandag:
verify

PR #165, opened 2019-04-12
add faketime to runner script

mdumandag:
add faketime to runner script to simulate an older system date for the ssl tests to past


mdumandag:
we bought a certificate from comodo that is valid for 2 years. There is no need for faketime anymore. I am closing this now

PR #175, opened 2019-07-03
Portable Serialization Multiversion Support Code Sample

yagmurdulger:
I added a code sample under examples > map to demonstrate the multiversion support of portable serialization.
It creates 3 versions of Employee class and configures 3 clients, each of them with a different version of the class. The code sample shows that the first 2 versions are appropriate for Portable multiversion support (addition of new field), while the third version fails to be compatible due to incompatible type change of one of its fields (field type change from int to String).

### DIFF
--- a/examples/map/map_portable_versioning_example.py
+++ b/examples/map/map_portable_versioning_example.py
@@ -3,6 +3,12 @@ import hazelcast
 
 from hazelcast.serialization.api import Portable
 
+# This sample code demonstrates multiversion support of Portable serialization.
+
+# With multiversion support, you can have two members that have different
+# versions of the same object, and Hazelcast will store both meta information and use the
+# correct one to serialize and deserialize portable objects depending on the member.
+
 # Default (version 1) Employee class.
 class Employee(Portable):
     FACTORY_ID = 666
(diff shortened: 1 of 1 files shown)


asimarslan:
verify

examples/map/map_portable_versioning_example.py
```diff
@@ -0,0 +1,186 @@
...
+
+    print('Map Size: {}'.format(my_map.size()))
+
+    # As clients with incompatible versions of the class try to access each other, a HazelcastSerializationError
+    # is raised (caused by a TypeError).
+    try:
+        # Client that has class with int type age field tries to read Employee3 object with String age field.
+        print(my_map.get(2))
+    except hazelcast.exception.HazelcastSerializationError as ex:
+        print("Failed due to: {}".format(ex.message))
```
mdumandag: `ex.message` statement throws error since `HazelcastSerializationError`does not have that field.  Just change it to `print("Failed due to: {}".format(ex))` to print the cause.

examples/map/map_portable_versioning_example.py
```diff
@@ -0,0 +1,186 @@
...
+
+    try:
+        # Client that has class with String type age field tries to read Employee object with int age field.
+        print(my_map3.get(0))
+    except hazelcast.exception.HazelcastSerializationError as ex:
+        print("Failed due to: {}".format(ex.message))
```
mdumandag: same as above

examples/map/map_portable_versioning_example.py
```diff
@@ -0,0 +1,186 @@
...
+
+    time.sleep(10)
```
mdumandag: Minor comment: I think that statement is not necessary, you can remove this and `import time` at the beginning of the script.

mdumandag:
verify

PR #176, opened 2019-07-03
Minor networking / serialization changes for better latency on single calls

tkountis:
- Replaced BUFFER_SIZE import with a local variable which respects the configuration (if exists)
- Removed the timeout of the asyncore loop to allow for the thread to finish faster, allowing other threads to make progress sooner
- Replaced the double list iteration on the Input#read_byte_array with a bytearray, similar to what was done here https://github.com/hazelcast/hazelcast-python-client/pull/174 but this version doesn't wrap the result, rather leaves it as pure `bytearray` which is **slightly concerning because it changes the type of the function**, but it gives a good bit of extra boost.

### DIFF
--- a/hazelcast/config.py
+++ b/hazelcast/config.py
@@ -706,6 +706,12 @@ class ClientProperties(object):
     Period in seconds to collect statistics.
     """
 
+    SERIALIZATION_INPUT_RETURNS_BYTEARRAY = ClientProperty("hazelcast.serialization.input.returns.bytearray", False)
+    """
+    Input#read_byte_array returns a List if property is False, otherwise it will return a byte-array.
+    Changing this to True, gives a considerable performance benefit.
+    """
+
     def __init__(self, properties):
         self._properties = properties
 
--- a/hazelcast/reactor.py
+++ b/hazelcast/reactor.py
@@ -42,7 +42,7 @@ class AsyncoreReactor(object):
         Future._threading_locals.is_reactor_thread = True
         while self._is_live:
             try:
-                asyncore.loop(count=1, timeout=0.1, map=self._map)
+                asyncore.loop(count=1, timeout=0.01, map=self._map)
                 self._check_timers()
             except select.error as err:
                 # TODO: parse error type to catch only error "9"
@@ -117,6 +117,7 @@ class AsyncoreReactor(object):
 
 class AsyncoreConnection(Connection, asyncore.dispatcher):
     sent_protocol_bytes = False
+    read_buffer_size = BUFFER_SIZE
 
     def __init__(self, map, address, connect_timeout, socket_options, connection_closed_callback,
                  message_callback, network_config, logger_extras=None):
@@ -135,6 +136,9 @@ class AsyncoreConnection(Connection, asyncore.dispatcher):
         self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, BUFFER_SIZE)
 
         for socket_option in socket_options:
+            if socket_option.option is socket.SO_RCVBUF:
+                self.read_buffer_size = socket_option.value
+
             self.socket.setsockopt(socket_option.level, socket_option.option, socket_option.value)
 
         self.connect(self._address)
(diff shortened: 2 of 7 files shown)


hazelcast/serialization/input.py
```diff
@@ -111,7 +111,7 @@ def read_byte_array(self):
         result = bytearray(length)
         if length > 0:
             self.read_into(result, 0, length)
-        return [x for x in result]
+        return result
```
tkountis: @mdumandag do you reckon this is a breaking change? I think the method should expect a bytearray input and produce bytearray output. if thats not the case, its the responsibility of the caller to address it, similar to what I did with the tests. We can add this in the release notes, but in my opinion its worth the pain, because it give good performance.

hazelcast/reactor.py
```diff
@@ -42,7 +42,7 @@ def _loop(self):
         Future._threading_locals.is_reactor_thread = True
         while self._is_live:
             try:
-                asyncore.loop(count=1, timeout=0.1, map=self._map)
```
mdumandag: I don't think we should set the timeout to 0. When it is 0, this becomes basically a busy loop with a select system call on each iteration. When the timeout is 0.1, 0.01 or 0.001 idle client uses %0-0.1 CPU on my laptop. When I set timeout to 0, CPU usage jumps to %18.  Also, setting the timeout to 0 slightly increased the latency on my machine. I think 0.1 or 0.01 would be a good value for the timeout.

hazelcast/serialization/input.py
mdumandag: I agree with you, this method should return a `bytearray` but doing it now is a breaking change since this method is in the public API. If we are going to release these fixes as 3.12.1 I think we should just return `list(result)`. @asimarslan what do you think ?

hazelcast/reactor.py
tkountis: You are right, I didn't see this on my desktop. I played with a few timeouts, and yes, sub-millis performs equally well. Changed it to 0.01.

hazelcast/serialization/input.py
tkountis: Thinking about it a bit more, how about if we keep the public API unchanged with a default arg? something like `as_list=True`. Updated the commit, have a look

asimarslan: @tkountis @mdumandag Can you validate the change with our backward compatibility tests? if the tests fail we can accept it. We never break backward compatibility for any enhancement for default behavior. We can only add a system property enabled feature that has breaking changes.

tests/serialization/identified_test.py
```diff
@@ -63,7 +63,7 @@ def read_data(self, inp):
         self.a_float = inp.read_float()
         self.a_double = inp.read_double()
 
-        self.bytes_ = inp.read_byte_array()
```
mdumandag: minor comment: these test changes should not be required anymore and maybe you can add an extra field to test the behaviour when the return value is bytearray

hazelcast/serialization/input.py
mdumandag: looks good to me

asimarslan: @tkountis I think we'll remove that parameter in 4.0 so putting a global config and keeping the public API clear make more sense to me. Maybe a new clientProperties config

tkountis: Done, @asimarslan @mdumandag please have another look

hazelcast/config.py
```diff
@@ -706,6 +706,12 @@ class ClientProperties(object):
     Period in seconds to collect statistics.
     """
 
+    SERIALIZATION_INPUT_RETURNS_BYTEARRAY = ClientProperty("hazelcast.serialization.input.returns.bytearray", False)
+    """
+    Input#read_byte_array returns a List if property is True, otherwise it will return a byte-array.
```
mdumandag: List if property is True -> List if property is False
Changing this to False -> Changing this to True

PR #181, opened 2019-07-24
Smart Listener Implementation

yagmurdulger:
This PR implements a listener service that enables the smart listening feature in smart-routing-enabled clients.
Smart listener service registers local-only listeners to each of the client's connections with the cluster and automatically registers/deregisters them as new connections are added/removed.


asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

hazelcast/listener.py
```diff
@@ -0,0 +1,163 @@
...
+        elapsed_millis = now_in_millis - start_millis
+        invocation_time_out_millis = self._invocation_service.invocation_timeout * 1000
+        timed_out = elapsed_millis > invocation_time_out_millis
+        if timed_out:
+            raise OperationTimeoutError\
+                ("Registering listeners is timed out. Last failed member: %s, Current time: %s, Start time: %s, "
+                 "Client invocation timeout: %s, Elapsed time: %s ms, Cause: %s", last_failed_member, now_in_millis,
+                 start_millis, invocation_time_out_millis, elapsed_millis, last_exception.args[0])
+        else:
+            self._invocation_service.invocation_retry_pause()  # sleep before next try
```
mdumandag: `invocation_retry_pause` is not a function

hazelcast/invocation.py
```diff
@@ -265,17 +192,13 @@ def _send(self, invocation, connection, ignore_heartbeat):
         try:
             connection.send_message(message)
         except IOError as e:
+            self._listener_service.remove_event_handler(correlation_id)
```
mdumandag: a check like `if invocation.event_handler is not None` as you did above may be better to not do the removal for non-listener related invocations

hazelcast/listener.py
```diff
@@ -0,0 +1,163 @@
...
+
+class ListenerService(object):
+    logger = logging.getLogger("HazelcastClient.ListenerService")
+
+    def __init__(self, client):
+        self._client = client
+        self._invocation_service = client.invoker
+        self.is_smart = client.config.network_config.smart_routing
+        self._logger_extras = {"client_name": client.name, "group_name": client.config.group_config.name}
+        self.active_registrations = {}  # Dict of user_registration_id, ListenerRegistration
```
mdumandag: `active_registrations` can be made private

hazelcast/listener.py
```diff
@@ -0,0 +1,163 @@
...
+        self.correlation_id = correlation_id
+
+
+class ListenerService(object):
+    logger = logging.getLogger("HazelcastClient.ListenerService")
+
+    def __init__(self, client):
+        self._client = client
+        self._invocation_service = client.invoker
+        self.is_smart = client.config.network_config.smart_routing
```
mdumandag: `is_smart ` can be made private

hazelcast/listener.py
```diff
@@ -0,0 +1,163 @@
...
+        def callback(f):
+            try:
+                response = f.result()
+                server_registration_id = listener_registration.decode_register_response(response)
+                correlation_id = registration_request.get_correlation_id()
+                registration = EventRegistration(server_registration_id, correlation_id)
+                registration_map[connection] = registration
+            except Exception as e:
+                if connection.live():
+                    self.logger.warning("Listener %s can not be added to a new connection: %s, reason: %s",
```
mdumandag: logger extras should be passed

hazelcast/listener.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def start(self):
+        self._client.connection_manager.add_listener(self.connection_added, self.connection_removed)
+
+    def handle_client_message(self, message):
+        correlation_id = message.get_correlation_id()
+        if correlation_id not in self._event_handlers:
+            self.logger.warning("Got event message with unknown correlation id: %s", message, extra=self._logger_extras)
+        event_handler = self._event_handlers.get(correlation_id)
+        event_handler(message)
```
mdumandag: what if event_handler is None ?

tests/listener_test.py
```diff
@@ -0,0 +1,86 @@
...
+        self.cluster = self.create_cluster(self.rc, None)
+        self.m1 = self.cluster.start_member()
+        self.m2 = self.cluster.start_member()
+        self.m3 = self.cluster.start_member()
+        self.client_config = ClientConfig()
+        self.collector = event_collector()
+
+    def tearDown(self):
+        self.rc.exit()
+        self.shutdown_all_clients()
```
mdumandag: it is better to shutdown clients before we close the members

tests/smart_listener_test.py
```diff
@@ -0,0 +1,78 @@
...
+    def tearDown(self):
+        self.shutdown_all_clients()
+
+    # -------------------------- test_local_only ----------------------- #
+    def test_list_smart_listener_local_only(self):
+        list = self.client.get_list(random_string()).blocking()
+        list.add_listener(item_added_func=self.collector)
+        list.add('item-value')
+        sleep(5)
+        self.assertEqual(1, len(self.collector.events))
```
mdumandag: better to use assertTrueEventually, then we can remove sleep call and test may finish earlier.

yagmurdulger: I purposefully avoided using assertTrueEventually in this case, because I want to make sure that I receive exactly 1 event and not more than that. I think assertTrueEventually lets the test pass as soon as you receive that 1 event, but I want to wait to see I don't receive the event from all of the listeners to make sure smart listener works local only.

hazelcast/listener.py
yagmurdulger: Although it looks like the only usages of is_smart is under ListenerService, Proxy also accesses that field:
self._is_smart = client.listener.is_smart
That's why I left it public.

mdumandag:
verify

hazelcast/listener.py
```diff
@@ -62,7 +63,7 @@ def time_out_or_sleep_before_next_try(self, start_millis, last_failed_member, la
                  "Client invocation timeout: %s, Elapsed time: %s ms, Cause: %s", last_failed_member, now_in_millis,
                  start_millis, invocation_time_out_millis, elapsed_millis, last_exception.args[0])
         else:
-            self._invocation_service.invocation_retry_pause()  # sleep before next try
+            sleep(self._invocation_service.invocation_retry_pause())  # sleep before next try
```
mdumandag: `sleep(self._invocation_service.invocation_retry_pause())`-> `sleep(self._invocation_service.invocation_retry_pause)`

mdumandag:
verify

PR #197, opened 2020-03-12
Implement distributed object listener and get_distributed_objects method

mdumandag:
This PR contains the code samples, documentation and implementation for the distributed object listener and get_distributed_objects method of the client.

Just like the Java client, call to the get_distributed_objects method clears the local instances of the destoyed proxies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@
   * [7.5. Distributed Events](#75-distributed-events)
     * [7.5.1. Cluster Events](#751-cluster-events)
       * [7.5.1.1. Listening for Member Events](#7511-listening-for-member-events)
-      * [7.5.1.2. Listening for Lifecycle Events](#7512-listening-for-lifecycle-events)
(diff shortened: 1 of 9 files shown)


filipArena:
Maybe support for async `get_distributed_objects()` makes sense - but I like it.

mdumandag:
Hi @filipArena , considering all the other blocking APIs on the HazelcastClient (get_map, get_list etc.), what is the use case of the async version of this API in your mind ? Maybe we can add `get_distributed_objects_async` or do something else. 

If you ask me, I am not really sure about the async version because I think it will look awkward since there is no async counterpart of the `get_data_structure` methods. We maybe consider adding those (or making them async by default for the upcoming 4.0 release in which we can do breaking changes).

@asimarslan What do you think ?

mdumandag:
@filipArena We discussed the possibility of the async API in the team and decided to continue with the sync version for now. But feel free to create an issue about it. We may consider it (along with the other changes I mentioned above) for the 4.0 release.

PR #200, opened 2020-04-03
Client4.0

buyalsky:
Protocol enhancements and cleanup


mdumandag:
Closing this since the client is compatible with the 4.0 clusters now

Issue #201, opened 2020-04-20
ReliableTopic for Python


burakcelebi:
Currently, the Python client lacks Hazelcast's ReliableTopic:
https://docs.hazelcast.org/docs/latest/manual/html-single/index.html#reliable-topic

Hazelcast's Reliable Topic uses the same [Topic](https://github.com/hazelcast/hazelcast-python-client#748-using-topic) interface as a regular topic. The main difference is that Reliable Topic is backed up by the [Ringbuffer](https://github.com/hazelcast/hazelcast-python-client#747-using-ringbuffer) data structure, a replicated but not partitioned data structure that stores its data in a ring-like structure.

buraksezer:
Hello,

I would like to take this issue.

burakcelebi:
Great, thanks @buraksezer! 

We are always here to help you :) cc: @asimarslan, @mdumandag

mdumandag:
Closing this, as the feature is implemented by

https://github.com/hazelcast/hazelcast-python-client/pull/395

PR #206, opened 2020-05-13
Initial ReliableTopic implementation: #201

buraksezer:
Hi team!

This is my initial PR for #201. I have several questions before completing the task.

**1:**
I need to add `config.set_property("hazelcast.serialization.input.returns.bytearray", True)` to serialize/deserialize messages. I set this property before running tests and sample programs. How do I prevent this?

**2:**
Test suite occasionally fails. See this [error messages](https://gist.github.com/buraksezer/e7cb4976c33030d13e2a365458bf64cd). I need to fix that issue before adding more tests. Because the missing tests are complicated and I don't understand the what the actual problem is.

**3:** 
How to add a `publisher_address`? The Go implementation doesn't use it.

**4:**
How to handle `publish_time`? It's Unix epoch now. We may need to use native Python dates here.

I extracted tests from the Go implementation. Currently some of the tests are missing. I'm going to add the following tests after fixing 2.

- TestReliableTopicProxy_ClientNotActiveError
- TestReliableTopicProxy_DistributedObjectDestroyedError
- TestReliableTopicProxy_DistributedObjectDestroyed
- TestReliableTopicProxy_Stale

I'm still iterating over the code. You can use `examples/reliable-topic/reliable_topic_example.py` for quick experments. Could you please share your ideas?

### DIFF
--- a/hazelcast/proxy/reliable_topic.py
+++ b/hazelcast/proxy/reliable_topic.py
@@ -251,7 +251,7 @@ class ReliableTopic(Proxy):
         Subscribes to this reliable topic. When someone publishes a message on this topic, on_message() method of
         ReliableTopicListener is called.
 
-        :param ReliableTopicListener: (Class), class to be used when a message is published.
+        :param reliable_topic_listener: (Class), class to be used when a message is published.
         :return: (str), a registration id which is used as a key to remove the listener.
         """
         if not isinstance(reliable_topic_listener, ReliableMessageListener):
--- a/hazelcast/proxy/reliable_topic.py
+++ b/hazelcast/proxy/reliable_topic.py
@@ -93,27 +93,36 @@ class _MessageListener(object):
         self._proxy.client.reactor.add_timer(0, self._next)
 
     def _handle_illegal_argument_error(self):
-        head_seq = self._proxy.ringbuffer.head_sequence().result()
-        self._proxy.logger.warning("MessageListener {} on topic {} requested a too large sequence. Jumping from old "
-                                   "sequence: {} to sequence: {}".format(self._id, self._proxy.name, self._sequence,
-                                                                         head_seq))
-        self._sequence = head_seq
-        self._next()
+        def on_response(res):
+            head_seq = res.result()
+            self._proxy.logger.warning("MessageListener {} on topic {} requested a too large sequence. Jumping from "
+                                       "old sequence: {} to sequence: {}".format(self._id, self._proxy.name,
+                                                                                 self._sequence,
+                                                                                 head_seq))
+            self._sequence = head_seq
+            self._next()
+
+        future = self._proxy.ringbuffer.head_sequence()
+        future.add_done_callback(on_response)
 
     def _handle_stale_sequence_error(self):
-        head_seq = self._proxy.ringbuffer.head_sequence().result()
-        if self._listener.is_loss_tolerant:
-            self._sequence = head_seq
-            self._proxy.logger.warning("Topic {} ran into a stale sequence. Jumping from old sequence {} to new "
-                                       "sequence {}".format(self._proxy.name, self._sequence, head_seq))
-            self._next()
-            return True
+        def on_response(res):
+            head_seq = res.result()
+            if self._listener.is_loss_tolerant:
+                self._proxy.logger.warning("Topic {} ran into a stale sequence. Jumping from old sequence {} to new "
+                                           "sequence {}".format(self._proxy.name, self._sequence, head_seq))
+                self._sequence = head_seq
+                self._next()
 
-        self._proxy.logger.warning(
-            "Terminating Message Listener: {} on topic: {}. Reason: The listener was too slow or the retention "
-            "period of the message has been violated. Head: {}, sequence: {}".format(self._id, self._proxy.name,
-                                                                                     head_seq, self._sequence))
-        return False
+            else:
+                self._proxy.logger.warning(
+                    "Terminating Message Listener: {} on topic: {}. Reason: The listener was too slow or the retention "
+                    "period of the message has been violated. Head: {}, sequence: {}".format(self._id, self._proxy.name,
+                                                                                             head_seq, self._sequence))
+                self._cancel_and_remove_listener()
+
+        future = self._proxy.ringbuffer.head_sequence()
+        future.add_done_callback(on_response)
 
     def _handle_operation_timeout_error(self):
         self._proxy.logger.info("Message Listener ", self._proxy.id, "on topic: ", self._proxy.name, " timed out. " +
@@ -126,8 +135,8 @@ class _MessageListener(object):
             self._handle_illegal_argument_error()
             return
         elif isinstance(exception, StaleSequenceError):
-            if self._handle_stale_sequence_error():
-                return
+            self._handle_stale_sequence_error()
+            return
         elif isinstance(exception, OperationTimeoutError):
             self._handle_operation_timeout_error()
             return
--- a/hazelcast/proxy/reliable_topic.py
+++ b/hazelcast/proxy/reliable_topic.py
@@ -1,12 +1,331 @@
-from hazelcast.proxy.base import Proxy
+import time
+import threading
+from uuid import uuid4
+
+from hazelcast.config import ReliableTopicConfig, TOPIC_OVERLOAD_POLICY
+from hazelcast.exception import IllegalArgumentError, TopicOverflowError, HazelcastInstanceNotActiveError, \
+    HazelcastClientNotActiveException, DistributedObjectDestroyedError, StaleSequenceError, OperationTimeoutError
+from hazelcast.proxy.base import Proxy, TopicMessage
+from hazelcast.proxy.ringbuffer import OVERFLOW_POLICY_FAIL, OVERFLOW_POLICY_OVERWRITE
+from hazelcast.serialization.reliable_topic import ReliableTopicMessage
+from hazelcast.util import current_time_in_millis
+from hazelcast.six.moves import queue
+
+_INITIAL_BACKOFF = 0.1
+_MAX_BACKOFF = 2
+
+
+class ReliableMessageListener(object):
+    def on_message(self, item):
+        """
+        Invoked when a message is received for the added reliable topic.
+
+        :param: message the message that is received for the added reliable topic
+        """
+        raise NotImplementedError
+
+    def retrieve_initial_sequence(self):
+        """
+        Retrieves the initial sequence from which this ReliableMessageListener
+        should start.
+
+        Return -1 if there is no initial sequence and you want to start
+        from the next published message.
+
+        If you intend to create a durable subscriber so you continue from where
+        you stopped the previous time, load the previous sequence and add 1.
+        If you don't add one, then you will be receiving the same message twice.
+
+        :return: (int), the initial sequence
+        """
+        return -1
+
+    def store_sequence(self, sequence):
+        """"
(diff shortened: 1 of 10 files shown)


mdumandag:
Hi @buraksezer , thank you so much for your efforts. It is great to see a community activity on the python client. I will be reviewing the PR in detail tomorrow, but I think it looks pretty good. For now, let me answer your questions.

1. Our `ObjectDataInput#read_byte_array` method returns list of byte-sized integers if the option you mentioned is not present. Honestly, I feel like the implementation should always return `bytearray` but, for backward compatibility reasons we cannot change it. The problem you have faced is in this line. https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/serialization/input.py#L150 . We are passing a list instead of bytearray. There is a missing check here. We should cast buff back into bytearray if `self._respect_bytearrays` is true. We will change the implementation of `read_byte_array` in 4.0 release, but for now we can use this.

2. I ran the test suite more than 10 times but couldn't reproduce the problem. Also, from the logs, I cannot tell anything useful. It seems like the member was not responding but I don't know why. I will look into it in more detail tomorrow.

3. You can simply pass None. It is there because the Java member side reliable topic proxy sends this data as a part of the message but clients does not. In the Java side, both member and the client use the same class so this option has to be present but can be None.

4. I am fine with returning Unix timestamp. It is easy to convert it to datetime object but I would like to leave this to user.

hazelcast/proxy/reliable_topic.py
```diff
@@ -1,12 +1,331 @@
...
+                sleep_time = _MAX_BACKOFF
+
+    def _add_or_fail(self, item):
+        seq_id = self.ringbuffer.add(item, overflow_policy=OVERFLOW_POLICY_FAIL).result()
+        if seq_id == -1:
+            raise TopicOverflowError("failed to publish message to topic: " + self.name)
+
     def publish(self, message):
-        raise NotImplementedError
+        """
```
mdumandag, line 287: This is a blocking API. In general the API provided in the 3.x versions are non-blocking apart from the proxy creations and listener registrations. So, I think `publish` should be non-blocking too.

hazelcast/proxy/reliable_topic.py
```diff
@@ -1,12 +1,331 @@
...
+    def start(self):
+        tail_seq = self._proxy.ringbuffer.tail_sequence()
+        initial_seq = self._listener.retrieve_initial_sequence()
+        if initial_seq == -1:
+            initial_seq = tail_seq.result() + 1
+        self._sequence = initial_seq
+        self._proxy.client.reactor.add_timer(0, self._next)
+
+    def _handle_illegal_argument_error(self):
+        head_seq = self._proxy.ringbuffer.head_sequence().result()
```
mdumandag, line 98: So, if we are at this line, we are on the reactor thread. If you call `.result()` on reactor thread, you will have a deadlock (because the reactor thread itself is the one sets the results of futures). Therefore, there is a check on the future objects that forces you to use `add_done_callback` on scenarios like this. I believe with the current test suite, we are not hitting this line.

hazelcast/proxy/reliable_topic.py
```diff
@@ -1,12 +1,331 @@
...
+        self._proxy.logger.warning("MessageListener {} on topic {} requested a too large sequence. Jumping from old "
+                                   "sequence: {} to sequence: {}".format(self._id, self._proxy.name, self._sequence,
+                                                                         head_seq))
+        self._sequence = head_seq
+        self._next()
+
+    def _handle_stale_sequence_error(self):
+        head_seq = self._proxy.ringbuffer.head_sequence().result()
```
mdumandag, line 106: Same as above

hazelcast/proxy/ringbuffer.py
```diff
@@ -157,7 +157,7 @@ def read_many(self, start_sequence, min_count, max_count):
         """
         check_not_negative(start_sequence, "sequence can't be smaller than 0")
         check_true(max_count >= min_count, "max count should be greater or equal to min count")
-        check_true(min_count <= self.capacity().result(), "min count should be smaller or equal to capacity")
+        #check_true(min_count <= self.capacity().result(), "min count should be smaller or equal to capacity")
```
asimarslan, line 160: The correct validation for this line is:
```
check_true(max_count <= self.capacity().result(), "max count should be smaller or equal to capacity")
```

asimarslan:
@mdumandag 
I think this PR should be sent to `3.12.z` (too)
`master` branch is for 4.x client development.
The contents are same for now but from a PM point of view we need it on 3.12.z branch.

hazelcast/proxy/reliable_topic.py
```diff
@@ -1,12 +1,337 @@
...
+            self._sequence = head_seq
+            self._next()
+
+        future = self._proxy.ringbuffer.head_sequence()
+        future.add_done_callback(on_response)
+
+    def _handle_stale_sequence_error(self):
+        def on_response(res):
+            head_seq = res.result()
+            if self._listener.is_loss_tolerant:
```
mdumandag, line 111: This should call the function. `if self._listener.is_loss_tolerant:` should be `if self._listener.is_loss_tolerant():`

tests/proxy/reliable_topic_test.py
```diff
@@ -0,0 +1,377 @@
...
+        stale_config.topic_overload_policy = TOPIC_OVERLOAD_POLICY.DISCARD_OLDEST
+        config.add_reliable_topic_config(stale_config)
+
+        self.client = hazelcast.HazelcastClient(self.configure_client(config))
+        self.reliable_topic = self.client.get_reliable_topic(random_string()).blocking()
+        self.registration_id = None
+
+    def tearDown(self):
+        if self.registration_id is not None:
+            self.reliable_topic.remove_listener(self.registration_id)
```
mdumandag, line 74: could you shutdown `self.client` after this line ?

tests/proxy/reliable_topic_test.py
```diff
@@ -0,0 +1,377 @@
...
+        self.cluster.start_member()
+
+        self.reliable_topic.publish("aa")
+
+        def assert_event():
+            self.assertEqual(len(collector.events), 1)
+            event = collector.events[0]
+            self.assertEqual(event.message, "aa")
+
+        self.assertTrueEventually(assert_event, 5)
```
mdumandag, line 348: You can shutdown the client_two by putting this line in try block and performing shutdown in the finally block.

tests/proxy/reliable_topic_test.py
```diff
@@ -0,0 +1,377 @@
...
+    @classmethod
+    def configure_cluster(cls):
+        path = os.path.abspath(__file__)
+        dir_path = os.path.dirname(path)
+        with open(os.path.join(dir_path, "hazelcast_topic.xml")) as f:
+            return f.read()
+
+    def setUp(self):
+        config = ClientConfig()
+        config.set_property("hazelcast.serialization.input.returns.bytearray", True)
```
mdumandag, line 50: As I mentioned before, you can remove this line after performing the small fix I mentioned here https://github.com/hazelcast/hazelcast-python-client/pull/206#issuecomment-628522527 (item 1)

hazelcast/proxy/ringbuffer.py
buraksezer: @asimarslan,  @mdumandag  Sorry for this extreme delay. I was overloaded by my day job and the other open source development stuff. Now I'm back to the Hazelcast community.  

I changed the validation line and added some logic to call `read_many` in `_MessageListener._next` function. Without this modification, `read_many` call in the ReliableTopic implementation throws the exception:

```
HazelcastClient.LifecycleService: INFO: [dev] [hz.client_1] (20200531 - 2df48a6) HazelcastClient is CONNECTED
HazelcastClient: INFO: [dev] [hz.client_1] Client started.
HazelcastClient.AsyncoreReactor: ERROR: [dev] [hz.client_1] Error in Reactor Thread
  File "/home/buraksezer/Projects/python/hazelcast-python-client/hazelcast/reactor.py", line 46, in _loop
    self._check_timers()
  File "/home/buraksezer/Projects/python/hazelcast-python-client/hazelcast/reactor.py", line 66, in _check_timers
    if timer.check_timer(now):
  File "/home/buraksezer/Projects/python/hazelcast-python-client/hazelcast/reactor.py", line 277, in check_timer
    self.timer_ended_cb()
  File "/home/buraksezer/Projects/python/hazelcast-python-client/hazelcast/proxy/reliable_topic.py", line 211, in _read_many
    future = self._proxy.ringbuffer.read_many(self._sequence, 1, self._proxy.config.read_batch_size)
  File "/home/buraksezer/Projects/python/hazelcast-python-client/hazelcast/proxy/ringbuffer.py", line 160, in read_many
    check_true(max_count <= self.capacity().result(), "max count should be smaller or equal to capacity")
  File "/home/buraksezer/Projects/python/hazelcast-python-client/hazelcast/future.py", line 58, in result
    self._reactor_check()
  File "/home/buraksezer/Projects/python/hazelcast-python-client/hazelcast/future.py", line 70, in _reactor_check
    "Synchronous result for incomplete operation must not be called from Reactor thread. "
RuntimeError: Synchronous result for incomplete operation must not be called from Reactor thread. Use add_done_callback instead.
```

It seems that the corrected validation (max, instead of min) breaks some tests in `ringbuffer` implementation.

mdumandag: No problem at all, we are happy to see you back.

I checked the line and the error. I also played with the version that has no capacity check but it seems there is no check on the server-side. So, when you have a ring buffer with 10 elements capacity, if you make a `read_many` request with `min_count` larger than that, the server never sends back a response (waits to get at least `min_count` items even if it is higher than its capacity).

So, I guess we have to have some kind of capacity check to fail on bad user requests. I checked your implementation but it seems that even if the capacity check fails, the future returned to the user is not resolved with that failure. I tweaked it a little bit to do that. What do you think about it? I did some tests and it seems it is working.

```python

        check_not_negative(start_sequence, "sequence can't be smaller than 0")
        check_true(max_count >= min_count, "max count should be greater or equal to min count")
        check_true(max_count < MAX_BATCH_SIZE, "max count can't be greater than %d" % MAX_BATCH_SIZE)

        future = Future()

        def set_result(result):
            try:
                future.set_result(result.result())
            except:
                future.set_exception(sys.exc_info()[1], sys.exc_info()[2])

        def check_capacity(capacity):
            try:
                capacity = capacity.result()
                check_true(min_count <= capacity, "min count: %d should be smaller or equal to capacity: %d"
                           % (min_count, capacity))
                f = self._encode_invoke(ringbuffer_read_many_codec, response_handler=self._read_many_response_handler,
                                        start_sequence=start_sequence, min_count=min_count,
                                        max_count=max_count, filter=None)
                f.add_done_callback(set_result)
            except:
                future.set_exception(sys.exc_info()[1], sys.exc_info()[2])

        self.capacity().add_done_callback(check_capacity)

        return future

```

yuce:
Hi @buraksezer, I hope everything's going great with you.

Thanks for your contribution! Unfortunately this PR fell out of sync with our main branch before the reviews were addressed. Would it be possible for you to resolve the conflicts and address the reviews?

buraksezer:
Hi @yuce 

I'm so sorry about that but I have no free time to complete the work.

yuce:
@buraksezer No problem, thanks for letting us now. Good luck!

mdumandag:
Closing this in favor of #395. @buraksezer thanks again for your efforts. We have used your commits as a base for the implementation.

PR #209, opened 2020-06-26
Update Hazelcast Cloud documentation and code sample

### LINKED ISSUE #207: Confusing chapter about Cloud Discovery
See this chapter: https://github.com/hazelcast/hazelcast-python-client/blob/v3.12.3/README.md#58-enabling-hazelcast-cloud-discovery

The very first sentence: _The purpose of Hazelcast Cloud Discovery is to provide the clients the means to use IP addresses provided by hazelcast orchestrator_

As a new user I am confused: _What is "hazelcast orchestrator" and why should I care?_
I don't think users care about such details. Users wants to connect to a cluster. That's all.

Then it also says:
_To be able to connect to the provided IP addresses, you should use secure TLS/SSL connection between the client and members. Therefore, you should enable the SSL configuration as described in the TLS/SSL for Hazelcast Python Client section._

I don't understand it. There is no _should_. If I enabled TLS in the Cloud console then I _must_ enable it on the client side too. If I didnt not enable TLS on a server-side then I must not enable it on a client side.

mdumandag:
We didn't updated Hazelcast Cloud section since we have implemented
it. Right now, Hazelcast Cloud does not use secure communication
by default. One have to enable encryption before starting the cluster
to use secure communication. Also, unlike the code sample we have
provided, Hazelcast Cloud does not use certificates signed by
trusted CA. It generates a private certificate pair. This commit
updates the code sample accordingly.

fixes #207

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@
   * [7.5. Distributed Events](#75-distributed-events)
     * [7.5.1. Cluster Events](#751-cluster-events)
       * [7.5.1.1. Listening for Member Events](#7511-listening-for-member-events)
-      * [7.5.1.2. Listenring for Distributed Object Events](#7512-listening-for-distributed-object-events)
+      * [7.5.1.2. Listening for Distributed Object Events](#7512-listening-for-distributed-object-events)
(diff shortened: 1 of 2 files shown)


mdumandag:
closing this, the changes on this PR is already on the master

PR #215, opened 2020-08-31
Implement map.set_ttl

### LINKED ISSUE #214: Missing functionality - map.set_ttl
It seems that map proxy has a missing functionality: `set_ttl`. It should be trivial to implement this. 

- We just need to generate codecs using the old protocol branch (https://github.com/hazelcast/hazelcast-client-protocol/tree/1.8.z) and copy the map_set_ttl_codec to the protocol directory.
- Implement the functionality (should be very similar to this https://github.com/hazelcast/hazelcast/blob/3.12.z/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java#L1425)
- Add documentation and tests

Reported on: https://stackoverflow.com/questions/63402015/setting-ttl-record-expiry-in-hazelcast-for-individual-entries-without-a-put/63402407

yemreinci:
Fixes #214.


hazelcast/proxy/map.py
```diff
@@ -725,6 +725,21 @@ def set(self, key, value, ttl=-1):
...
+        value is valid starting from the time this operation is invoked, not since the time the entry was created.
+        If the entry does not exist or is already expired, this call has no effect.
+
+        :param key: (object), the key of the map entry.
+        :param ttl: (int), maximum time for this entry to stay in the map (0 means infinite,
+            negative means map config default)
+        """
+        check_not_none(key, "key can't be None")
+        check_not_none(ttl, "ttl can't be None")
+        key_data = self._to_data(key)
```
mdumandag, line 740: I see that in the Java client, near cache invalidates the key once you call the `set_ttl` on it. We should also do that here.

Basically,  you should move the `return self._encode_invoke_on_key(map_set_ttl_codec, key_data, key=key_data, ttl=to_millis(ttl))` part to a method called `_set_internal` and in the near cache implementation, invalidate the key. Something like this https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/proxy/map.py#L1002

mdumandag (approved):
Thanks @yemreinci 

Could you also backport this PR to the 3.12.z branch? The content of the backport should be identical to this one

PR #218, opened 2020-09-09
Add UuidSerializer and improve serialization service and its test suite

mdumandag:
Depends on #217 

This PR implements the UuidSerializer as a default serializer.
With this change, objects of `uuid.UUID` type will be serialized using this serializer and `java.util.UUID` objects will be converted to that type when received from the server.

Apart from that, this PR includes some other changes.
- Objects of `bytearray` type will now be serialized with `ByteArraySerializer`
- Some small improvements on the hot path of the `to_data` function.
- Removing faulty range checks from `IntegerSerializer` and `LongSerializer`. Formerly, we were throwing an exception for `-1 << 31` and `-1 << 63` values. However, they are valid as `Integer` or `Long`. Now, we are relying on the `struct` package to throw exceptions on out-of-range values.
- Faulty constant definitions are fixed. (`MAX_SHORT`, `MAX_INT`, `MAX_LONG`) Unused constant definitions are removed. (`MAX_FLOAT32`, `MIN_FLOAT32`)
- Unreachable methods (like `JavaClassSerializer.write_data`, `LinkedListSerializer.write_data` etc.) are commented out. 
- `BigDecimalSerializer` is removed.
- Implementation of `DatetimeSerializer` is fixed and made compatible with `java.util.Date`.
- The serialization test suite is significantly improved to cover all serializers in various scenarios like `set`/`get` from the client, `set` from the client, `get` from the remote controller, `set` on the server, and `get` from the client.


mdumandag:
verify

sancar (approved):

PR #224, opened 2020-09-30
CP Subsytem - Atomic Reference

mdumandag:
For the AtomicReference, implementation, code samples, documentation and
tests are added. It also includes some minor docstring fixes regarding
the lists in some places and some missing checks and documentation fixes
for the AtomicLong.

depends on #223

### DIFF
--- a/examples/cp/atomic_reference_example.py
+++ b/examples/cp/atomic_reference_example.py
@@ -1,0 +1,17 @@
+import hazelcast
+
+client = hazelcast.HazelcastClient()
+
+my_ref = client.cp_subsystem.get_atomic_reference("my-ref").blocking()
+my_ref.set(42)
+
+value = my_ref.get()
+print("Value:", value)
+
+result = my_ref.compare_and_set(42, "value")
+print("CAS result:", result)
+
+final_value = my_ref.get()
+print("Final value:", final_value)
+
+client.shutdown()
--- a/hazelcast/proxy/cp/atomic_long.py
+++ b/hazelcast/proxy/cp/atomic_long.py
@@ -2,6 +2,7 @@ from hazelcast.protocol.codec import atomic_long_add_and_get_codec, atomic_long_
     atomic_long_get_codec, atomic_long_get_and_add_codec, atomic_long_get_and_set_codec, atomic_long_alter_codec, \
     atomic_long_apply_codec
 from hazelcast.proxy.cp import BaseCPProxy
+from hazelcast.util import check_not_none
 
 
 class AtomicLong(BaseCPProxy):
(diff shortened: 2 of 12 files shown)


examples/cp/atomic_reference_example.py
```diff
@@ -0,0 +1,17 @@
+import hazelcast
```
puzpuzpuz, line 1: nit: does it make sense to add an org web sample for AtomicReference?

hazelcast/proxy/cp/atomic_long.py
```diff
@@ -145,13 +146,18 @@ def alter(self, function):
...
-                The function applied to the currently stored value.
+                The function that alters the currently stored value.
 
         Returns:
             hazelcast.future.Future[None]:
         """
+        check_not_none(function, "Function cannot be None")
         function_data = self._to_data(function)
         codec = atomic_long_alter_codec
+        # 1 means return the new value.
```
puzpuzpuz, line 157: That's a nice addition. I had to look at the codec and server-side code to understand what `1`/`0` means. 👍

hazelcast/proxy/cp/atomic_long.py
```diff
@@ -145,13 +146,18 @@ def alter(self, function):
 
         Args:
             function (hazelcast.serialization.api.Portable or hazelcast.serialization.api.IdentifiedDataSerializable):
-                The function applied to the currently stored value.
+                The function that alters the currently stored value.
 
         Returns:
             hazelcast.future.Future[None]:
         """
+        check_not_none(function, "Function cannot be None")
```
puzpuzpuz, line 154: A potentially stupid question. Don't we need to validate number inputs in other methods, like `new_value` in `get_and_set`?

puzpuzpuz (approved):
LGTM. Left a few questions

examples/cp/atomic_reference_example.py
mdumandag: I thought about adding it but I saw that we don't have the sample for the AtomicReference in 3.12 series. So, I didn't add it but I think we should discuss the structure of the org-website samples within the team and decide on a set of samples. We can add it afterward. What do you think about it?

hazelcast/proxy/cp/atomic_long.py
mdumandag: In fact, it makes a lot of sense. We don't have type checks in most of the codebase, we can improve that incrementally but It makes sense to add them for the new features. I also added simple tests for invalid input cases

examples/cp/atomic_reference_example.py
puzpuzpuz: Yes, we can add those samples later, after an in-team discussion.

mdumandag:
Thanks for the review Andrey

PR #225, opened 2020-10-01
CP Subsystem - CountDownLatch

mdumandag:
Adds the implementation, documentation, code samples, and test of
CP CountDownLatch.

Also, the invocation service is edited to return `OperationTimeoutError`
instead of `HazelcastTimeoutError` in case of invocation timeouts
as the Java client does.

depends on #224

### DIFF
--- a/tests/proxy/cp/count_down_latch_test.py
+++ b/tests/proxy/cp/count_down_latch_test.py
@@ -1,0 +1,153 @@
+import time
+from threading import Thread
+
+from hazelcast.errors import DistributedObjectDestroyedError, OperationTimeoutError
+from hazelcast.future import ImmediateExceptionFuture
+from hazelcast.util import AtomicInteger
+from tests.proxy.cp import CPTestCase
+from tests.util import random_string
+
+inf = 2 ** 31 - 1
+
+
+class CountDownLatchTest(CPTestCase):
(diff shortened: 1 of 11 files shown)


mdumandag:
verify

tests/proxy/cp/count_down_latch_test.py
```diff
@@ -0,0 +1,168 @@
...
+        self.assertEqual(0, latch.get_count())
+        latch.try_set_count(10)
+        self.assertEqual(10, latch.get_count())
+
+    def test_try_set_count(self):
+        latch = self._get_latch()
+        self.assertTrue(latch.try_set_count(3))
+        self.assertEqual(3, latch.get_count())
+
+    def test_try_set_count_when_count_is_already_set(self):
```
puzpuzpuz, line 135: nit: does it make sense include a test that verifies that `try_set_count` succeeds once the count goes to zero?

puzpuzpuz (approved):
LGTM

tests/proxy/cp/count_down_latch_test.py
mdumandag: Good idea. I added the extra test case.

mdumandag:
Thanks for the review Andrey, I am merging this PR

PR #229, opened 2020-10-15
Follow advised logging practices

mdumandag:
Contrary to what https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
advises, we were defining our own StreamHandler with custom formatter and
filter. It is strongly advised for libraries to not do that. We are following that
advice now. In short, the changes done are described below.

- Use module-level loggers with the module name as the logger name. It is described
as a good convention in the https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial.
- Do not add a Handler with custom formatter and filter by default.
- Do not print the latest commit id in the Lifecycle service as it requires reading a file
during the client startup. Instead, the client version is printed. To make test failure
reports easier to follow, the latest commit id is printed for the test logs.
- For tests, a single configuration is added to the root logger in the init file.
That way, all tests will reuse that.
- Do not pass logger extras anymore as they don't make sense without a custom formatter or
filter. We may provide utility implementations for a formatter and a filter and pass
extras even if they won't be used most of the time. I chose not to pass them, but that
can be discussed.
- Remove `logging_level` and `logging_config` confiugration options.
- Update README.


README.md
```diff
@@ -2581,148 +2569,86 @@ After enabling the client statistics, you can monitor your clients using Hazelca
...
-For example, setting the logging level to `logging.DEBUG` will cause all the logging messages that are equal or higher than the `logging.DEBUG` in terms of severity to be emitted by your logger.
+2020-10-16 13:31:35,624 - hazelcast.client - INFO - Client started
+2020-10-16 13:31:35,624 - hazelcast.lifecycle - INFO - HazelcastClient 4.0.0 is SHUTTING_DOWN
+2020-10-16 13:31:35,624 - hazelcast.connection - INFO - Removed connection to Address(host=127.0.0.1, port=5701):7682c357-3bec-4841-b330-6f9ae0c08253, connection: Connection(id=0, live=False, remote_address=Address(host=172.17.0.2, port=5701))
+2020-10-16 13:31:35,624 - hazelcast.lifecycle - INFO - HazelcastClient 4.0.0 is DISCONNECTED
+2020-10-16 13:31:35,634 - hazelcast.lifecycle - INFO - HazelcastClient 4.0.0 is SHUTDOWN
+```
 
-By default, the logging level is set to `logging.INFO`.
+A handy alternative to above example would be configuring the root logger using the ``logging.basicConfig()`` utility method. Beware that,
```
puzpuzpuz, line 2616:
```suggestion
A handy alternative to the above example would be configuring the root logger using the ``logging.basicConfig()`` utility method. Beware that,
```

README.md
```diff
@@ -2581,148 +2569,86 @@ After enabling the client statistics, you can monitor your clients using Hazelca
...
-``` 
+INFO:hazelcast.client:Client started
+INFO:hazelcast.lifecycle:HazelcastClient 4.0.0 is SHUTTING_DOWN
+INFO:hazelcast.connection:Removed connection to Address(host=127.0.0.1, port=5701):7682c357-3bec-4841-b330-6f9ae0c08253, connection: Connection(id=0, live=False, remote_address=Address(host=172.17.0.2, port=5701))
+INFO:hazelcast.lifecycle:HazelcastClient 4.0.0 is DISCONNECTED
+INFO:hazelcast.lifecycle:HazelcastClient 4.0.0 is SHUTDOWN
+```
 
-To learn more about the `logging` module and its capabilities, please see the [logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html) and [documentation](https://docs.python.org/3/library/logging.html) of the `logging` module.
+To learn more about the `logging` package and its capabilities, please see the [logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html) and [documentation](https://docs.python.org/3/library/logging.html) of the `logging` package.
```
puzpuzpuz, line 2651: nit: does it make sense to also put links for Python 2 somewhere here, as we support both versions?

hazelcast/cluster.py
```diff
@@ -214,15 +214,15 @@ def _apply_new_state_and_fire_events(self, current, snapshot):
                     try:
                         handler(removed_member)
                     except:
-                        self.logger.exception("Exception in membership lister", extra=self._logger_extras)
+                        _logger.exception("Exception in membership lister")
```
puzpuzpuz, line 217:
```suggestion
                        _logger.exception("Exception in membership listener")
```

hazelcast/cluster.py
```diff
@@ -214,15 +214,15 @@ def _apply_new_state_and_fire_events(self, current, snapshot):
...
 
         for added_member in additions:
             for handler, _ in six.itervalues(self._listeners):
                 if handler:
                     try:
                         handler(added_member)
                     except:
-                        self.logger.exception("Exception in membership lister", extra=self._logger_extras)
+                        _logger.exception("Exception in membership lister")
```
puzpuzpuz, line 225:
```suggestion
                        _logger.exception("Exception in membership listener")
```

hazelcast/config.py
```diff
@@ -1,8 +1,3 @@
-"""
-Hazelcast Client Configuration module contains configuration classes and various constants required to create a ClientConfig.
```
puzpuzpuz, line 2: Don't we need this doc comment here? Or it was redundant?

hazelcast/cluster.py
puzpuzpuz, line 217: A stupid question: don't we need to log some information about the raised exception here and on L224?

hazelcast/invocation.py
```diff
@@ -122,21 +122,21 @@ def shutdown(self):
     def _invoke_on_partition_owner(self, invocation, partition_id):
         owner_uuid = self._partition_service.get_partition_owner(partition_id)
         if not owner_uuid:
-            self.logger.debug("Partition owner is not assigned yet", extra=self._logger_extras)
+            _logger.debug("Partition owner is not assigned yet")
```
puzpuzpuz, line 125: Don't we need `if _logger.isEnabledFor(logging.DEBUG)` check here and around other debug messages in this script?

hazelcast/reactor.py
```diff
@@ -203,7 +203,7 @@ def __init__(self, dispatcher_map, connection_manager, connection_id, address,
 
     def handle_connect(self):
         self.start_time = time.time()
-        self.logger.debug("Connected to %s", self.connected_address, extra=self._logger_extras)
+        _logger.debug("Connected to %s", self.connected_address)
```
puzpuzpuz, line 206: nit: the same consideration on debug level check. The same consideration applies to debug logging in other scripts changed by this PR, so I won't duplicate this comment anymore.

puzpuzpuz (approved):
Looks like a good improvement 👍

hazelcast/config.py
mdumandag: I don't think we need this. It contains no useful information and the `ClientConfig` is not there anymore. The configuration classes also do not exist anymore and, their documentation is moved to the HazelcastClient's constructor.

hazelcast/cluster.py
mdumandag: `.exception()` does that automatically. It inspects the current exception stack (and finds the one we caught) and logs it

README.md
mdumandag: I don't think it is necessary. Python2 users might switch to their version using the dropdown menu on the navbar in that link.

hazelcast/invocation.py
mdumandag: We don't have this check in most places. It is necessary only if we pass an argument that results in some computation. For example, for
```python
_logger.debug("xyz %s", costly_call())
```
the check makes sense. But for
```python
_logger.debug("xyz %s", some_reference)
```
it is not necessary because the logging package is smart enough to do string interpolation lazily only if the logger is enabled for that log level.

But thanks to your comment, I saw that we were using lazy interpolation only on some places, there were lots of non-lazy interpolations of the form
```python
_logger.debug("xyz %s" % some_reference)
```
I changed them with the correct approach on https://github.com/hazelcast/hazelcast-python-client/pull/229/commits/238e963318db7f6ae74592294c47ac68d7c46cd5

README.md
puzpuzpuz: OK, makes sense. Let's leave it as is then.

hazelcast/config.py
puzpuzpuz: Makes sense

PR #232, opened 2020-10-16
Implement PagingPredicate

mdumandag:
This is the port of #212
to the 4.0 branch. However, there are some differences from that PR.

- Documentation of the feature is a little bit different since the client
side sorting is removed.
- Tests are more-or-less ported directly with a few additions for
input validations. However, the above PR was using `six.assertCounEqual`
for assertions which checks equality but not order. For paging predicate
tests, it makes more sense to check ordering.
- Client side codecs and holder classes for PagingPredicates and AnchorDataLists
are added.
- Some public methods, like `get_anchor`, `set_anchor`, `get_nearest_anchor_entry`
is removed. `get_anchor` is defined on the public API on the Java side but I think
it does not provide any functionality to the user.

Also, some missing `six.moves.range` imports are added.

Protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/355

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@
       * [7.7.1.2. Querying by Combining Predicates with AND, OR, NOT](#7712-querying-by-combining-predicates-with-and-or-not)
       * [7.7.1.3. Querying with SQL](#7713-querying-with-sql)
       * [7.7.1.4. Querying with JSON Strings](#7714-querying-with-json-strings)
+      * [7.7.1.5. Filtering with Paging Predicates](#7715-filtering-with-paging-predicates)
   * [7.8. Performance](#78-performance)
     * [7.8.1. Near Cache](#781-near-cache)
       * [7.8.1.1. Configuring Near Cache](#7811-configuring-near-cache)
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -2422,6 +2423,60 @@ You can configure this using `metadata-policy` element for the map configuration
...
+Therefore, you need to register the serialization factories you use on all the members on which the paging predicates are used.
+See the [Adding User Library to CLASSPATH](#1212-adding-user-library-to-classpath) section for more details.
+
+In the example code below:
+
+- The `is_greater_than_or_equal_to` predicate gets values from the `students` map. 
+This predicate has a filter to retrieve the objects with an `age` greater than or equal to `18`.
+
+- Then a `PagingPredicate` is constructed in which the page size is `5`, so that there are five objects in each page. 
+The first time the values are called creates the first page.
```
puzpuzpuz, line 2445: nit: I'm not sure I understand what's meant by this sentence. It's better to re-write it in a clearer way.

README.md
```diff
@@ -2422,6 +2423,60 @@ You can configure this using `metadata-policy` element for the map configuration
...
+
+- It gets subsequent pages with the `next_page()` method of `PagingPredicate` and querying the map again with the updated `PagingPredicate`.
```
puzpuzpuz, line 2447:
```suggestion
- Finally, the subsequent page is fetched by calling the `next_page()` method of `PagingPredicate` and querying the map again with the updated `PagingPredicate`.
```

README.md
```diff
@@ -2422,6 +2423,60 @@ You can configure this using `metadata-policy` element for the map configuration
...
+
+# Set up next page
+predicate.next_page()
+
+# Retrieve next page
+values = m.values(predicate)
+```
+
+If a comparator is not specified for `PagingPredicate`, but you want to get a collection of keys or values page by page, 
+this collection must be an instance of `Comparable` (i.e., it must implement `java.lang.Comparable` on the member side).
```
puzpuzpuz, line 2470: nit: maybe it's more precise to say that keys/values should implement `Comparable`, not their collections?

README.md
```diff
@@ -2422,6 +2423,60 @@ You can configure this using `metadata-policy` element for the map configuration
...
+...
+
+# Set up next page
+predicate.next_page()
+
+# Retrieve next page
+values = m.values(predicate)
+```
+
+If a comparator is not specified for `PagingPredicate`, but you want to get a collection of keys or values page by page,
```
puzpuzpuz, line 2469: nit: does it make sense to stress the fact that the `If a comparator is not specified for `PagingPredicate`` part speaks of the cluster-side class? Python users may be confused otherwise.

README.md
```diff
@@ -2422,6 +2423,60 @@ You can configure this using `metadata-policy` element for the map configuration
...
+If a comparator is not specified for `PagingPredicate`, but you want to get a collection of keys or values page by page, 
+this collection must be an instance of `Comparable` (i.e., it must implement `java.lang.Comparable` on the member side). 
+Otherwise, paging fails with an exception from the server. Luckily, a lot of types implement the `Comparable`
+interface by [default](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html), including the primitive types,
+so, you may use values of types `int`, `float`, `str` etc. in paging without specifying a comparator on the Python client.
+
+You can also access a specific page more easily by setting the `predicate.page` attribute before making the remote call.
+This way, if you make a query for the hundredth page, for example, it gets all `100` pages at once instead of reaching the hundredth page one by one using the `next_page()` method. 
+
+`PagingPredicate`, also known as Order & Limit, is not supported in Transactional Context.
```
puzpuzpuz, line 2478: nit: does it make sense to convert this one into a note (smth like `> **NOTE: foo bar**` style used in Node.js client)? It seems to be a side note which falls of the previous context.

hazelcast/util.py
```diff
@@ -281,3 +281,18 @@ def next(self):
             return None
         idx = random.randrange(0, len(members))
         return members[idx]
+
+
+class IterationType:
+    """To differentiate users selection on result collection on map-wide
+    operations like values, keySet, query etc.
```
puzpuzpuz, line 288: nit: does it make sense to change the list of operations to `entry_set, key_set, values` to make things more Python client specific?

puzpuzpuz (approved):
LGTM

hazelcast/util.py
mdumandag: Good idea, changed that

README.md
mdumandag: Added it

README.md
mdumandag: Good idea, changed that part

README.md
mdumandag: Changed

README.md
mdumandag: I think it is related to the ``values`` method used in the code sample, rewrite that part

mdumandag:
@puzpuzpuz , I agree with the comments you have made. I copy-pasted almost all of them from the Java reference manual, I will also send a pull request there.

README.md
mdumandag: I believe this part is meant for the cases like `paging(predicate, 5, None)` or `paging(predicate, 5)` when the comparator is not specified on the client-side. Did I get your question wrong?

puzpuzpuz: Probably, I misread this part. Thanks for the clarification. Resolving this one.

mdumandag:
Thanks for the quick review

PR #235, opened 2020-10-26
Fix class definition registrations with the same class id

### LINKED ISSUE #199: Explicitly registering Portable ClassDefinitions fails with multiple PortableFactories and overlapping classIds
see following issue description 
https://github.com/hazelcast/hazelcast/issues/16820

Java fix
https://github.com/hazelcast/hazelcast/pull/16831

mdumandag:
When class definitions for two different classes with the same class id
but different factory ids are registered, we were throwing an error
indicating that there are duplicate registrations. However, we should
allow such cases.

Apart from the fix for that, the PR also includes a test case for
null portable serialization.

Closes #199

### DIFF
--- a/hazelcast/serialization/service.py
+++ b/hazelcast/serialization/service.py
@@ -78,24 +78,40 @@ class SerializationServiceV1(BaseSerializationService):
         self._registry.safe_register_serializer(self._registry._python_serializer)
 
     def register_class_definitions(self, class_definitions, check_error):
-        class_defs = dict()
+        factories = dict()
         for cd in class_definitions:
-            if cd in class_defs:
-                raise HazelcastSerializationError("Duplicate registration found for class-id: %s" % cd.class_id)
-            class_defs[cd.class_id] = cd
-        for cd in class_definitions:
-            self.register_class_definition(cd, class_defs, check_error)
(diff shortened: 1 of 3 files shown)


hazelcast/serialization/service.py
```diff
@@ -78,24 +78,40 @@ def _register_constant_serializers(self):
...
+            if class_id in class_defs:
+                raise HazelcastSerializationError("Duplicate registration found for class-id: %s" % class_id)
+            class_defs[class_id] = cd
+
         for cd in class_definitions:
-            self.register_class_definition(cd, class_defs, check_error)
+            self.register_class_definition(cd, factories, check_error)
 
-    def register_class_definition(self, cd, class_defs, check_error):
+    def register_class_definition(self, cd, factories, check_error):
```
puzpuzpuz, line 97: nit: it may be a good idea to rename `cd` into `class_definition` to keep the same level of verbosity for parameter names as in `register_class_definitions(self, class_definitions, check_error)`

puzpuzpuz (approved):

hazelcast/serialization/service.py
mdumandag: Renamed it

PR #240, opened 2020-10-28
Fix the race on the timers

mdumandag:
Formerly, we were doing a logic like the following: Check the
timer in front of the priority queue of timers. If it is expired,
call the callback and pop it from the queue. However, this logic
is flawed because the timers queue was accessed from multiple threads.
So, it may be the case that, we check the timer in front of the queue,
run its callback, then, some other thread adds another timer to the
queue concurrently and it is put in front of the queue. Now, when the
reactor thread pops the item in front of the queue, it will be the newly
added timer, before its callback being executed.

To solve this, we use a double buffering approach. A thead-safe queue
is used to store newly added timers which can be efficiently modified
on both ends on multiple threads. Then, the reactor thread periodically
pops items from that queue in FIFO order and maintains a min heap
using a list with the help of heappush and heappop functions. Only if
the min heap has some elements, timers are popped from that and executed.

Also, the unneeded `timer_cancelled_cb` is field is removed. We were
removing the timer from the queue after canceling it. That was the
only usage of this field. We do the same with the new approach,
if it is canceled, it will return `True` on `check_timer` call without
executing its `timer_ended_cb` and will be removed from the heap.

### DIFF
--- a/hazelcast/reactor.py
+++ b/hazelcast/reactor.py
@@ -9,6 +9,7 @@ import time
 
 from collections import deque
 from functools import total_ordering
+from heapq import heappush, heappop
 
 from hazelcast import six
 from hazelcast.config import SSLProtocol
@@ -16,7 +17,6 @@ from hazelcast.connection import Connection
 from hazelcast.core import Address
 from hazelcast.errors import HazelcastError
 from hazelcast.future import Future
-from hazelcast.six.moves import queue
 
 try:
     import ssl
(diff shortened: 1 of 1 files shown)


hazelcast/reactor.py
```diff
@@ -55,28 +56,36 @@ def _loop(self):
...
+
+        if timers:
+            now = time.time()
+            while timers:
+                timer = timers[0][1]
+                if timer.check_timer(now):
+                    heappop(timers)
+                else:
+                    # Timer in the root of the min heap is not expired
+                    # Therefore, there should not be any expired
```
puzpuzpuz, line 82:
```suggestion
                    # Therefore, there should be no expired
```

puzpuzpuz (approved):
LGTM

PR #241, opened 2020-11-10
Optimize murmur hash calculation

mdumandag:
We were always passing a bytes-like object to the murmur hash function
but, we were copying it from the HEAP_DATA_OVERHEAD(8) offset, which
can be huge. I removed that copy and tuned the offsets according to our
HEAP_DATA_OVERHEAD.

Another parameter of the function was the data size. For valid Data
objects, this is always equal to the `len(buffer) - HEAP_DATA_OVERHEAD`.
This parameter is removed and this length is calculated inside the
murmur hash function now.

Also, inlined the _fmix function call and removed the unnecessary
method calls on hash_to_index function.

### DIFF
--- a/hazelcast/hash.py
+++ b/hazelcast/hash.py
@@ -11,8 +11,8 @@ def murmur_hash3_x86_32(data):
     Returns:
         int: Calculated hash value.
     """
-    length = len(data) - 8  # Heap data overhead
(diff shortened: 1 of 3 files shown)


puzpuzpuz (approved):

mdumandag:
Thanks for the review Andrey

PR #242, opened 2020-11-10
Make writes non-blocking

mdumandag:
Formerly, writes were checking the size of the write queue, and
if there are no items in it, were writing directly to the socket.
That is against the nature of the non-blocking API we provide.
Also, it restains us from optimizing the writes for non-blocking
usage (multiple client messages can be concatenated into single
buffer before writing to the socket to decrease the number of
send calls).

To solve that problem, connections are adding buffers into the
write queue. Reactor thread pops elements from that queue and
writes into the socket. (The optimization I mentioned above
can be implemented later). Of course, this approach will kill the
performance of blocking usage since the reactor thread will be
in sleep most of the time on asyncore.loop call. We need to wake
the reactor thread each time we want to write something if it is
not awake.

To do this, a pipe is added to the reactor's dispatchers map. Each
time we want to write something, we check whether or not it is awake
and if not, we write a single byte into the pipe. That wakes the
reactor thread and makes it process the message we added to its
write queue. The pipe works great on UNIX, but pipes are not
selectable or pollable on Windows. We fall back into sockets in case
the client is running on Windows.

On the reactor, we try to create the loop, check if the added waker
(pipe or socket) works or not. If it does, we use the wakable loop.
If not, we fall back to the loop that does busy waiting.


hazelcast/reactor.py
```diff
@@ -228,17 +433,21 @@ def handle_read(self):
...
+        while True:
             try:
                 data = self._write_queue.popleft()
             except IndexError:
                 return
+
             sent = self.send(data)
             self.last_write_time = time.time()
             self.sent_protocol_bytes = True
             if sent < len(data):
```
puzpuzpuz, line 445: I don't the place where we set `O_NONBLOCK` for connection sockets, so they should operate in blocking mode. This means that we may block the reactor thread on `send` and `recv`. But I'm not sure if it's a good idea to start using non-blocking mode for sockets. WDYT?

hazelcast/reactor.py
```diff
@@ -228,17 +433,21 @@ def handle_read(self):
             reader.process()
 
     def handle_write(self):
-        with self._write_lock:
```
puzpuzpuz, line 231: Leaving a comment for L426 `data = self.recv(self.read_buffer_size)` here. Don't we need to have a `data == b''` check for the closed connection case after this call?

hazelcast/reactor.py
```diff
@@ -101,33 +258,81 @@ def shutdown(self):
...
+    def __init__(self):
+        self.map = {}
+        loop = None
+        try:
+            loop = _WakeableLoop(self.map)
+            loop.check_loop()
+        except:
+            if loop:
+                loop.shutdown()
+                loop = _BusyWaitLoop(self.map)
```
puzpuzpuz, line 319: nit: it might be a good idea to log a warn message here, as this implementation may be sub-optimal in certain scenarios.

puzpuzpuz (approved):
LGTM in general. Left some comments for further discussion, but there is nothing major

hazelcast/reactor.py
mdumandag: Setting timeout to 0 makes sockets non-blocking and we are doing that(L413, L118, L119). I didn't change that part for connection sockets, but looking at the code, errors thrown by the non-blocking sockets are ignored in the `handle_error` (also didn't change that part). I didn't hear complaints from users on that part of the client, so I am assuming there should be no problem

hazelcast/reactor.py
mdumandag: This is handled by the asyncore internally. If the received data has 0 length, the connection on our side will also be closed.

hazelcast/reactor.py
puzpuzpuz: Makes sense. I've overlooked that part, thanks.

Considering this, the old implementation which was writing in the socket immediately in case of an empty queue wasn't blocking, right?

hazelcast/reactor.py
puzpuzpuz: I see, thanks.

hazelcast/reactor.py
mdumandag: Maybe the blocking was not the correct word here, but it was waiting to write it to the socket. Thinking about it again, the time it takes for send(basically copying this buffer to kernel buffer) shouldn't take long, but I believe the approach in this PR is a little bit better and allows future improvements. What do you think?

puzpuzpuz: Yes, I still see value in this approach if we coalesce messages before writing them. But it might be a good idea to implement a PoC and benchmark it. This way we'll be certain that this is the right direction. WDYT?

mdumandag: I did this, I see up to %20 throughput improvement for balanced use cases(more for write-heavy scenarios). The code is almost ready, needs just a little bit of polishing. I will try to make it into the beta release. At worst, we can ship this in the final release.

hazelcast/reactor.py
```diff
@@ -83,13 +200,53 @@ def _check_timers(self):
...
+        self.wake_loop()
+        assert self.waker.awake
+        self.run_loop()
+        assert not self.waker.awake
+
+    def run_loop(self):
+        asyncore.loop(timeout=0.1, use_poll=True, map=self._map, count=1)
+
+    def wake_loop(self):
+        self.waker.wake()
```
puzpuzpuz, line 249: Does it make sense to check that the caller thread is different from the reactor thread? It doesn't make any sense to wake `poll` when we're already on the reactor thread.

hazelcast/reactor.py
puzpuzpuz: Thanks for the clarification! Resolving this one.

hazelcast/reactor.py
mdumandag: If we are in the reactor thread and calling this, we are already awake(not waiting on the poll call), so we are just writing a single byte to a pipe(or socket), which is almost no-op

puzpuzpuz: I don't think that it's not almost no-op. In case when there are no new external (unrelated with the waker pipe) events, writing into the waker will mean that the next `poll` will immediately exit (and we'll have to do two system calls instead of one), while it should lay dormant. I can see that Tornado has this check: https://github.com/tornadoweb/tornado/blob/branch5.1/tornado/ioloop.py#L1128

But this is definitely a nit, so it's up to you to implement the check or not.

mdumandag: Added that

tests/reactor_test.py
```diff
@@ -0,0 +1,254 @@
...
+            original_close()
+
+        loop.waker.close = assertion
+
+        loop.shutdown()
+
+    @parameterized.expand(LOOP_CLASSES)
+    def test_check_loop(self, _, cls):
+        loop = cls({})
+        loop.check_loop()
```
puzpuzpuz, line 69: nit: I'd leave a comment that we rely on `assert`s placed in the method implementation here to make things a bit more straight-forward for future readers.

mdumandag: Wrote a comment explaining that

PR #243, opened 2020-11-10
Remove logging on sending each client message

mdumandag:
It is not really useful for the end user to see the each message
that clients send. It might be useful for us on some occasions,
so I left it as a commented line. Commenting it resulted in
approx. %1-2 throughput increase.


puzpuzpuz (approved):

puzpuzpuz:
I'd prefer to remove the line completely or log the message only when debug level is on, but I'm fine with the proposed change too.

PR #244, opened 2020-11-11
Serialization service improvements

mdumandag:
See the individual commit messages for the explanation.

### DIFF
--- a/hazelcast/serialization/data.py
+++ b/hazelcast/serialization/data.py
@@ -16,8 +16,8 @@ class Data(object):
     It stores binary form of an object serialized by serialization service.
     """
 
-    def __init__(self, buff=None):
-        self._buffer = buff
+    def __init__(self, buf):
+        self._buffer = buf
 
     def to_bytes(self):
         """Returns byte array representation of internal binary format.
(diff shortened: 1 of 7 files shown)


hazelcast/serialization/data.py
```diff
@@ -76,17 +77,6 @@ def is_portable(self):
...
 
-    def has_partition_hash(self):
-        """Determines whether this ``Data`` has partition hash or not.
-        
-        Returns:
-            bool: ``True`` if ``Data`` has partition hash, ``False`` otherwise.
-
-        """
-        return self._buffer is not None \
-               and len(self._buffer) >= HEAP_DATA_OVERHEAD \
```
puzpuzpuz, line 87: nit: it seems that we don't have this check anymore. Previous implementation was returning `self.hash_code()` when `len(self._buffer) < HEAP_DATA_OVERHEAD`. But I don't think this situation can be faced in the current implementation. WDYT?

puzpuzpuz (approved):
LGTM

hazelcast/serialization/data.py
mdumandag: Yes, I checked that part. We try to get partition hash only when the Data is created from the ObjectDataOutput, which will always be bigger than that

mdumandag:
Thanks for the review Andrey

PR #245, opened 2020-11-18
Add SOAK test

mdumandag:
Adds the script that runs the SOAK test, helper scripts to start
members and clients and sample server configuration file.


tests/soak_test/map_soak_test.py
```diff
@@ -0,0 +1,200 @@
...
+
+    logging.info("Soak test operations are starting!")
+    logging.info("* " * 20 + "\n")
+
+    test_failed = [False]
+
+    def run(stats):
+        end_time = time.time() + hour_limit * 60 * 60
+        while time.time() < end_time:
+            if test_failed[0]:
```
puzpuzpuz, line 108: Is this check thread safe? Or to be more precise, does Python guarantee that the updated list value will be published to all threads?

tests/soak_test/hazelcast.xml
```diff
@@ -0,0 +1,48 @@
...
+        <ssl enabled="false"/>
+        <socket-interceptor enabled="false"/>
+    </network>
+    <serialization>
+        <data-serializable-factories>
+            <data-serializable-factory factory-id="66">com.hazelcast.client.test.IdentifiedFactory
+            </data-serializable-factory>
+        </data-serializable-factories>
+    </serialization>
+</hazelcast>
```
puzpuzpuz, line 48: nit: a new line is missing here.

tests/soak_test/start_client.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+PID=$$
+ADDRESSES=$1
+HOUR=${2:-48.0}
+
+mkdir -p "client_logs"
+
+python map_soak_test.py --hour "$HOUR" --addresses "$ADDRESSES" --log "client_logs/client-${PID}" &
```
puzpuzpuz, line 9: nit: a new line is missing here.

tests/soak_test/start_server.sh
```diff
@@ -0,0 +1,13 @@
...
+HAZELCAST_VERSION="4.0"
+PID=$$
+
+mkdir -p "server_logs"
+
+CLASSPATH="hazelcast-${HAZELCAST_VERSION}.jar:hazelcast-${HAZELCAST_TEST_VERSION}-tests.jar"
+CMD_CONFIGS="-Dhazelcast.multicast.group=224.206.1.1 -Djava.net.preferIPv4Stack=true"
+java "${CMD_CONFIGS}" -cp ${CLASSPATH} \
+    com.hazelcast.core.server.HazelcastMemberStarter \
+    > server_logs/hazelcast-${HAZELCAST_VERSION}-${PID}-out.log 2>server_logs/hazelcast-${HAZELCAST_VERSION}-${PID}-err.log &
```
puzpuzpuz, line 13: nit: a new line is missing here.

tests/soak_test/start_server.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+HAZELCAST_TEST_VERSION="4.0"
+HAZELCAST_VERSION="4.0"
+PID=$$
+
+mkdir -p "server_logs"
+
+CLASSPATH="hazelcast-${HAZELCAST_VERSION}.jar:hazelcast-${HAZELCAST_TEST_VERSION}-tests.jar"
+CMD_CONFIGS="-Dhazelcast.multicast.group=224.206.1.1 -Djava.net.preferIPv4Stack=true"
```
puzpuzpuz, line 10: nit: we already define multicast group in `hazelcast.xml`, so probably `-Dhazelcast.multicast.group=224.206.1.1` could be omitted.

tests/soak_test/start_server.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+HAZELCAST_TEST_VERSION="4.0"
```
puzpuzpuz, line 3: nit: maybe it's better to use the latest available 4.0.z patch?

puzpuzpuz (approved):

tests/soak_test/map_soak_test.py
mdumandag: Well, Python does not have a memory model, so these details are not specified anywhere. 

But, CPython has GIL, so I guess every update should be visible to the thread that will hold the GIL next time. It is mentioned here slightly https://www.python.org/dev/peps/pep-0583/#rationale

puzpuzpuz: BTW are we going to be any safer if we used one of built-in collection classes instead? But probably, such details aren't critical for a soak test.

mdumandag: Do you mean the funny usage of `test_failed[0]`? If it is, this is related to fact that you cannot edit a variable defined in the encapsulating scope from the closure. (There is `nonlocal` keyword in Python3 for this case, but not in Python2). So, I have to define something that can be modified inside the closure and picked the `list` since it is the simplest alternative.

puzpuzpuz: I see. OK, let's leave it this way then.

PR #248, opened 2020-11-20
Make ListenerService return Future

mdumandag:
Listener service was waiting for the result of the future
in a blocking way and, returning the registration id string.

This is not in line with our other APIs in which we provide
non-blocking API.

Also, the slight behaviour change applied to Java client is
ported to Python client. https://github.com/hazelcast/hazelcast/pull/17646

### DIFF
--- a/hazelcast/invocation.py
+++ b/hazelcast/invocation.py
@@ -261,7 +261,7 @@ class InvocationService(object):
         self._listener_service.register_listener(request,
                                                  codec.decode_response,
                                                  lambda reg_id: None,
-                                                 lambda m: codec.handle(m, self._backup_event_handler))
+                                                 lambda m: codec.handle(m, self._backup_event_handler)).result()
 
     def _backup_event_handler(self, correlation_id):
         invocation = self._pending.get(correlation_id, None)
--- a/hazelcast/proxy/map.py
+++ b/hazelcast/proxy/map.py
@@ -82,7 +82,7 @@ class Map(Proxy):
             loaded_func (function): Function to be called when an entry is loaded from a map loader.
 
         Returns:
-            str: A registration id which is used as a key to remove the listener.
+            hazelcast.future.Future[str]: A registration id which is used as a key to remove the listener.
         """
         flags = get_entry_listener_flags(ADDED=added_func, REMOVED=removed_func, UPDATED=updated_func,
                                          EVICTED=evicted_func, EXPIRED=expired_func, EVICT_ALL=evict_all_func,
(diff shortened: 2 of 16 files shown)


hazelcast/proxy/map.py
```diff
@@ -1246,15 +1246,12 @@ def _on_destroy(self):
...
-                lambda reg_id: map_remove_entry_listener_codec.encode_request(self.name, reg_id),
-                lambda m: codec.handle(m, self._handle_invalidation, self._handle_batch_invalidation))
-        except:
-            pass
+        codec = map_add_near_cache_invalidation_listener_codec
+        request = codec.encode_request(self.name, EntryEventType.INVALIDATION, self._is_smart)
+        self._invalidation_listener_id = self._register_listener(
+            request, lambda r: codec.decode_response(r),
+            lambda reg_id: map_remove_entry_listener_codec.encode_request(self.name, reg_id),
+            lambda m: codec.handle(m, self._handle_invalidation, self._handle_batch_invalidation)).result()
```
puzpuzpuz, line 1254: Is it ok that `_add_near_cache_invalidation_listener` is still blocking? I didn't check how it's used further, so that's why I'm asking.

hazelcast/invocation.py
```diff
@@ -261,7 +261,7 @@ def _register_backup_listener(self):
         self._listener_service.register_listener(request,
                                                  codec.decode_response,
                                                  lambda reg_id: None,
-                                                 lambda m: codec.handle(m, self._backup_event_handler))
+                                                 lambda m: codec.handle(m, self._backup_event_handler)).result()
```
puzpuzpuz, line 264: Is it ok that this call is blocking?

puzpuzpuz (approved):

hazelcast/proxy/map.py
mdumandag: This is called on the constructor of the near cached map if the invalidate on change option is true. I believe we should wait for this to complete. This is not a problem because `.get_XXX` methods on the Hazelcast client are blocking too. This was like this long time, so I didn't want to change that

hazelcast/invocation.py
mdumandag: `HazelcastClient()` call is also blocking. So, I think that shouldn't be a problem

puzpuzpuz: I see. Thanks for the clarification.

hazelcast/proxy/map.py
puzpuzpuz: I see. That's weird that we don't have non-blocking `get_XXX` calls, but that weirdness is also present in Java client. Thanks for the clarification.

mdumandag:

PR #259, opened 2020-12-08
Add serialization binary compatibility tests

### LINKED ISSUE #256: Add binary compatibility test for serialization
We should add binary compatibility tests for serialization using https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/nio/serialization/compatibility/BinaryCompatibilityFileGenerator.java

mdumandag:
Added tests that verify binary generated by Java can be deserialized
from the Python client, produces the same binary output when serialized by
the Python client and the Python client can serialize/deserialize
the reference objects used in these tests.

Also added documentation about the binary file. There are some items
we need to skip and change to make the size of the file smaller.

See the other commits for other fixes.

Closes #256


tests/serialization/binary_compatibility/README.md
```diff
@@ -0,0 +1,226 @@
+The serialization binary compatibility file is autogenerated using
```
puzpuzpuz, line 1: Nice documentation! It'll be certain valuable in future when the test needs to be updated.

puzpuzpuz (approved):

mdumandag:

PR #262, opened 2020-12-11
Fix setting configured socket options

mdumandag:
We forget to update the code that sets socket options while doing #219.

The code now handles the given socket options tuple correctly. Also,
added some tests that verifies this.


puzpuzpuz (approved):

mdumandag:

PR #264, opened 2020-12-14
Coalesce multiple client messages if possible before writing

mdumandag:
This is an optimization for the non-blocking usage where there are
multiple in-flight messages. We now coalesce them into a single buffer
and try to write it to the wire.

Before coming up with this solution, I tried a couple of different ones.
Here is the list of the main ones, and why they didn't work out:

- Having a single, pre-allocated bytearray per connection, and copying
message bytes into it. Even with the usage of `memoryview`s, slicing and
copying were so expensive.

- Implementing framed outbound client messages, and writing frame bytes
into a write buffer in the reactor thread. That resulted in approx. %5
throughput increase for balanced workloads.

This way of implementation results in approx. %10 throughput increase for
balanced workloads.

### DIFF
--- a/hazelcast/reactor.py
+++ b/hazelcast/reactor.py
@@ -1,5 +1,6 @@
 import asyncore
 import errno
+import io
 import logging
 import os
 import select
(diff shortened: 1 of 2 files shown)


puzpuzpuz (approved):

mdumandag:

PR #271, opened 2020-12-21
Capture the address in the loop of connect_to_all_members timer

mdumandag:
There was a bug in the implementation of the connect_to_all_members
timer. In the loop, we were not capturing the address variable
correctly. As recommended in https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result
we are now binding the variable while creating the closure.

### DIFF
--- a/hazelcast/connection.py
+++ b/hazelcast/connection.py
@@ -255,7 +255,12 @@ class ConnectionManager(object):
                         break
 
                     if not self.get_connection(member.uuid):
-                        self._get_or_connect(address).add_done_callback(lambda f: connecting_addresses.discard(address))
+                        # Bind the address to the value
+                        # in this loop iteration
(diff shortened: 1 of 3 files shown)


puzpuzpuz (approved):
Nice catch! Is there a chance that we have similar problems in other places within the code base?

mdumandag:
Thanks for the review. I checked the codebase last Friday but let me check it again

mdumandag:
@puzpuzpuz I have found two such bugs, could you take a look again?

puzpuzpuz (approved):

puzpuzpuz:
> @puzpuzpuz I have found two such bugs, could you take a look again?

LGTM. Thanks for catching all of issues!

PR #273, opened 2020-12-21
Add ability to use enum literals in config

### LINKED ISSUE #228: Allow using literals for enum-like configs
For example, to configure the reconnect mode, one has to do the following.

```python
import hazelcast

from hazelcast.config import ReconnectMode


client = hazelcast.HazelcastClient(
    reconnect_mode=ReconnectMode.OFF,
)
```

We can get rid of the import and extra typing required by also allowing configurations like below
```python
import hazelcast

client = hazelcast.HazelcastClient(
    reconnect_mode="OFF",
)
```

mdumandag:
This PR adds the ability to use enum literals in the config
to make it easier to set these options, without importing anything.

When a value for enum is set, we first check if it is an actual
value for an enum(i.e., 0 for SSLProtocol.SSLv2). If it is, we
directly use that. If not, and the given configuration value is
a string, we try to get the enum member associated with that name.
If it matches to any member(i.e., SSLv2 for SSLProtocol.SSLv2),
we get the value associated with it and set it as the value.

If none of the above works, we throw TypeError as before.

Closes #228

### DIFF
--- a/docs/securing_client_connection.rst
+++ b/docs/securing_client_connection.rst
@@ -51,6 +51,8 @@ configuration and then go over the configuration options one by one:
         ssl_certfile="/home/hazelcast/certfile.pem",
         ssl_keyfile="/home/hazelcast/keyfile.pem",
         ssl_password="keyfile-password",
+        # You can also set this to "TLSv1_3"
+        # without importing anything.
         ssl_protocol=SSLProtocol.TLSv1_3,
         ssl_ciphers="DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA"
     )
@@ -188,6 +190,8 @@ follows:
     from hazelcast.config import SSLProtocol
 
     client = hazelcast.HazelcastClient(
+        # You can also set this to "TLSv1_3"
+        # without importing anything.
         ssl_protocol=SSLProtocol.TLSv1_3
     )
 
--- a/docs/securing_client_connection.rst
+++ b/docs/securing_client_connection.rst
@@ -190,8 +190,6 @@ follows:
     from hazelcast.config import SSLProtocol
 
     client = hazelcast.HazelcastClient(
-        # You can also set this to "TLSv1_3"
-        # without importing anything.
         ssl_protocol=SSLProtocol.TLSv1_3
     )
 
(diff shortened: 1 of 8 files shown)


docs/securing_client_connection.rst
```diff
@@ -188,6 +190,8 @@ follows:
     from hazelcast.config import SSLProtocol
 
     client = hazelcast.HazelcastClient(
+        # You can also set this to "TLSv1_3"
```
puzpuzpuz, line 193: nit: this fact was already mentioned above, so this comment could be removed for the sake of brevity

docs/using_python_client_with_hazelcast_imdg.rst
```diff
@@ -1745,6 +1745,8 @@ configuration, its default value is used.
                 "invalidate_on_change": True,
                 "time_to_live": 60,
                 "max_idle": 30,
+                # You can also set these to "OBJECT"
+                # or "LRU" without importing anything.
```
puzpuzpuz, line 1749:
```suggestion
                # and "LRU" without importing anything.
```

hazelcast/config.py
```diff
@@ -216,21 +216,15 @@ def unique_key(self):
 
     @unique_key.setter
     def unique_key(self, value):
-        if get_attr_name(QueryConstants, value):
-            self._unique_key = value
-        else:
-            raise TypeError("unique_key must be of type QueryConstants")
-    
+        try_to_set_enum_value(self, value, QueryConstants, "_unique_key")
```
puzpuzpuz, line 219: nit: I'd prefer `try_to_set_enum_value` to have no side effects and just convert the `int|string` into the expected `string`. This way there will be no need to pass the stringly typed field name (i.e. `"_unique_key"`) into the function. As another side effect, the function can be tested in an easier way. It could look something like the following:
```python
self._unique_key = try_to_get_enum_value(self, value, QueryConstants)
```

puzpuzpuz (approved):

hazelcast/config.py
mdumandag: Thanks for the suggestion. Changed to something similar with no side effects

puzpuzpuz: Thanks for doing the change. The only improvement I can think of is renaming `try_to_set_enum_value` into `try_to_get_enum_value` or something similar. That's because the function no longer sets the value - it only converts it.

mdumandag: Oh, I missed that. Updated it

puzpuzpuz: Thanks

mdumandag:

PR #274, opened 2020-12-23
Refactor the reactor connection, write and read logic

mdumandag:
We have a couple of problematic parts in the core of the reactor.

The first one is that we run periodic tasks that may open up connections,
such as the timer in the connection manager that tries to open
connections to members in the memberlist. In the timer, we do not
wait for authentication, etc. so it is not blocking. But if the address
we try to connect is unreachable(for example, in case of the pod restarts in k8s)
socket.connect call(which we do in the constructor of the AsyncoreConnection)
will block as long as the connection timeout which is 5s by default.
That means that, if you have 3 members in the memberlist, and they all become
unreachable to the client, reactor thread may be blocked for 3 * 5 = 15s
in the timer, while waiting for the socket.connect to return.

To overcome this, we now connect to sockets non-blockingly (with a 0s timeout)
and spawn a timer that will run after connection timeout seconds, that will
close the connection, if it is not connected at that time. If a problem occurs
before the timer, we close the connection and cancel the timer.

Apart from that, we also had a problem with the handle_read/write logic.
After some research, I saw that non-blocking sockets may throw
EAGAIN and EWOULDBLOCK errors, which should be retried. Also, if the SSL is
enabled, it may throw SSL_ERROR_WANT_WRITE/READ. See
https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html.

Formerly, if the handle_write/read throws EAGAIN or EDEADLK, we were not
closing the connection, but also not retrying the write correctly. For example,
we pop a message from the write queue, try to send it, get EAGAIN, go to handle_error,
simply ignore the error and go on. But, the popped message was never appended back
to the write queue. Therefore, the error handling logic is moved inside the
handle_write/read, and proper message retry is implemented.

Regarding the EDEADLK, I couldn't find a proper reason to retry it. It was
added in
72c6537.
I only saw this notable library doing it,
https://github.com/docker/docker-py/blob/master/docker/utils/socket.py#L28
not sure about it, but to not change the client's behavior on some edge cases,
I am adding it to the retryable exceptions.


hazelcast/discovery.py
```diff
@@ -1,15 +1,11 @@
...
 import logging
+import ssl
 
 from hazelcast.errors import HazelcastCertificationError
 from hazelcast.core import AddressHelper
 from hazelcast.six.moves import http_client
 
-try:
-    import ssl
-except ImportError:
```
puzpuzpuz, line 10: What was the reason for this check? `ssl` seems to be available in both Python 2 and 3.

hazelcast/reactor.py
```diff
@@ -38,6 +33,22 @@
 
 _logger = logging.getLogger(__name__)
 
+# We should retry receiving/sending the message in case of these errors
+# EAGAIN or EWOULDBLOCK: The read/write would block
+# EDEADLK: Was added before, retrying it just to make sure that
+#   client behaves the same on some edge cases.
+# SSL_ERROR_WANT_READ/WRITE: The socket could not satisfy the
+#   needs of the SSL_read/write. During the negotiation process
+#   SSL_read/write may also wants to write/read data, hence may also
```
puzpuzpuz, line 42:
```suggestion
#   SSL_read/write may also want to write/read data, hence may also
```

hazelcast/reactor.py
```diff
@@ -38,6 +33,22 @@
 
 _logger = logging.getLogger(__name__)
 
+# We should retry receiving/sending the message in case of these errors
+# EAGAIN or EWOULDBLOCK: The read/write would block
```
puzpuzpuz, line 37: nit: [manpages](https://man7.org/linux/man-pages/man3/errno.3.html) say that EAGAIN in general means "Resource temporarily unavailable", but it might be (but not always) the same as EWOULDBLOCK.

puzpuzpuz (approved):
Nice catches!

hazelcast/reactor.py
mdumandag: Separated the explanations for them

hazelcast/discovery.py
mdumandag: As you said, it was an unnecessary check

mdumandag:

hazelcast/discovery.py
puzpuzpuz: Thanks for the confirmation.

PR #281, opened 2021-01-05
[3.12.z] - Fix remove_distributed_object_listener test

### LINKED ISSUE #280: [3.12.z] tests.proxy.distributed_objects_test.DistributedObjectsTest.test_remove_distributed_object_listener
http://jenkins.hazelcast.com/job/python2.7-3.12.z-lab-master/10/testReport/junit/tests.proxy.distributed_objects_test/DistributedObjectsTest/test_remove_distributed_object_listener/

Probably, we should backport #261 to 3.12.z branch

mdumandag:
In the test, we were creating a map, and then immediately remove
the listener. This works most of the time, but sometimes, event
messages may arrive later, after we got a response from the get_map
request. If we remove the listener after the get_map response, but
before the late event, we will miss the event.

Test is modified to wait for the created event, then removing the
listener, then sleeping for a while and checking that there are no
further events.

Backport of #261
Closes #280

### DIFF
--- a/tests/proxy/distributed_objects_test.py
+++ b/tests/proxy/distributed_objects_test.py
@@ -1,3 +1,5 @@
+import time
+
 import hazelcast
 from hazelcast.core import DistributedObjectEventType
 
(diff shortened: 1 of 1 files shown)


sancar (approved):

mdumandag:
Thanks for the reviews @sancar

PR #297, opened 2021-01-13
Set the cliet version to 4.0

mdumandag:
Set the client version to 4.0 and development status to
production/stable after completing the release process.


puzpuzpuz (approved):
Congrats on the release!

PR #298, opened 2021-01-14
README.rst: Code Samples

burakcelebi:
PR closed by Hazelcast automation as no activity (>3 months). Please reopen with comments, if necessary. Thank you for using Hazelcast and your valuable contributions


yuce (approved):

mdumandag:
A bit too late, but I will try to review anyways :smile: 

I don't think we should add code samples to that section. It is not a channel where the user can ask questions. I feel like, If we want to include a link to code samples, we need to do it elsewhere

srknzl:
I think it is helpful to include the link to the code samples. I would delete the first sentence `You can use the following channels for your questions and development/usage issues:`

and leave the rest like as it is in this PR

srknzl:
Actually a link to the examples is included in the `getting started` document. Both options is fine for me then.

PR #299, opened 2021-01-19
Rewrite scripts

mdumandag:
Rewrite the scripts that download the remote controller and run tests
in Python instead of bash or powershell.

The motivation is to reduce duplication. We had to write scripts
for bash and powershell to make it easier to test the client in
different operating systems.

That won't be necessary with a single Python script like this.

Also, added the ability to skip enterprise tests to the test runner
for the users who do not have the HAZELCAST_ENTERPRISE_KEY in
their environment variables.


puzpuzpuz (approved):
That's a good change. Created a similar issue for Node.js client: https://github.com/hazelcast/hazelcast-nodejs-client/issues/752

mdumandag:
Thanks for the review Andrey

PR #343, opened 2021-01-27
Cleanup resources when the connection attempt fails immediately

mdumandag:
When the socket.connect fails immediately, we were not removing
the connection from the dispatchers map and closing the
in memory write buffer.


yuce:
@mdumandag Could you add a test for this fix?

yuce (approved):
Thanks, LGTM

PR #346, opened 2021-01-28
Set the cluster connection timeout to infinite by default

### LINKED ISSUE #306: [TRACKING ISSUE] Set cluster connect timeout to infinite by default

**Set cluster connect timeout to infinite by default**

See https://github.com/hazelcast/hazelcast/pull/18094 for details.

---

After an internal discussion, we have decided to set the cluster
connect timeout to infinite by default. Also, the multiplier
for the sleep time between consecutive connect attempts is changed
to `1.05` from `1.0`. The new sleep times between connection
attempts can be seen in the following graph. https://www.desmos.com/calculator/hl7rnbk4gf
(x-axis: attempt number, y-axis: sleep time in seconds)

Also, updated some logging statements.

Note that, this is a breaking change for the users who rely on the client to stop
trying to connect cluster after some time, but we decided

### LINKED ISSUE #335: [TRACKING ISSUE] Increase default cluster timeout to 2 minutes

**Increase default cluster timeout to 2 minutes**

See https://github.com/hazelcast/hazelcast/pull/18019 for details.

---

The old default tiemout was 20 seconds.
The change is done for environments like Kubernetes
where an Ip resolution/system restart could take
around a minute. We don't want a client to shutdown
in such cases, and we also want this to be out-of-the-box
behaviour.

Related config API

clientConfig.getConnectionStrategyConfig()
.getConnectionRetryConfig().getClusterConnectTimeoutMillis();

fixes https://github.com/hazelcast/hazelcast/issues/17904

mdumandag:
We decided to set the default cluster connection timeout to
infinite to have a better user experience out-of-the-box
for most of the users/use cases. Note that, this is a breaking
change for the users who rely on the client to shutdown after
some time. Affected users might set the `cluster_connect_timeout`
to a finite value.

Also, updated the default value of the `retry_multiplier` to back off
more after the client tries to connect to the target cluster for
some time.

Closes #335 
Closes #306

### DIFF
--- a/docs/client_connection_strategy.rst
+++ b/docs/client_connection_strategy.rst
@@ -46,10 +46,10 @@ starting and reconnecting modes.
 Configuring Client Connection Retry
 -----------------------------------
 
-When the client is disconnected from the cluster, it searches for new
-connections to reconnect. You can configure the frequency of the
-reconnection attempts and client shutdown behavior using the argumentes
-below.
+When the client is disconnected from the cluster or trying to connect
+to a one for the first time, it searches for new connections. You can
+configure the frequency of the connection attempts and the client
+shutdown behavior using the arguments below.
 
 .. code:: python
 
@@ -69,13 +69,14 @@ The following are configuration element descriptions:
 - ``retry_max_backoff``: Specifies the upper limit for the backoff in
   seconds. Its default value is ``30``. It must be non-negative.
 - ``retry_multiplier``: Factor to multiply the backoff after a failed
-  retry. Its default value is ``1``. It must be greater than or equal
+  retry. Its default value is ``1.05``. It must be greater than or equal
   to ``1``.
 - ``retry_jitter``: Specifies by how much to randomize backoffs. Its
   default value is ``0``. It must be in range ``0`` to ``1``.
 - ``cluster_connect_timeout``: Timeout value in seconds for the client
   to give up to connect to the current cluster. Its default value is
-  ``120``.
+  ``-1``. For the default value, client will not stop trying to connect
+  to the target cluster. (infinite timeout)
 
 A pseudo-code is as follows:
 
@@ -86,6 +87,7 @@ A pseudo-code is as follows:
     while (try_connect(connection_timeout)) != SUCCESS) {
         if (get_current_time() - begin_time >= CLUSTER_CONNECT_TIMEOUT) {
             // Give up to connecting to the current cluster and switch to another if exists.
+            // For the default values, CLUSTER_CONNECT_TIMEOUT is infinite.
         }
         sleep(current_backoff + uniform_random(-JITTER * current_backoff, JITTER * current_backoff))
         current_backoff = min(current_backoff * MULTIPLIER, MAX_BACKOFF)
--- a/docs/client_connection_strategy.rst
+++ b/docs/client_connection_strategy.rst
@@ -74,7 +74,7 @@ The following are configuration element descriptions:
 - ``retry_jitter``: Specifies by how much to randomize backoffs. Its
   default value is ``0``. It must be in range ``0`` to ``1``.
 - ``cluster_connect_timeout``: Timeout value in seconds for the client
-  to give up to connect to the current cluster. Its default value is
+  to give up connecting to the cluster. Its default value is
   ``-1``. For the default value, client will not stop trying to connect
   to the target cluster. (infinite timeout)
 
--- a/docs/client_connection_strategy.rst
+++ b/docs/client_connection_strategy.rst
@@ -46,10 +46,9 @@ starting and reconnecting modes.
 Configuring Client Connection Retry
 -----------------------------------
 
-When the client is disconnected from the cluster or trying to connect
-to a one for the first time, it searches for new connections. You can
-configure the frequency of the connection attempts and the client
-shutdown behavior using the arguments below.
+The client searches for new connections when it is trying to connect
+to the cluster. Both the frequency of connection attempts and the
+client shutdown behavior can be configured using the arguments below.
 
 .. code:: python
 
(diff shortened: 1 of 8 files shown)


docs/client_connection_strategy.rst
```diff
@@ -69,13 +69,14 @@ The following are configuration element descriptions:
 - ``retry_max_backoff``: Specifies the upper limit for the backoff in
   seconds. Its default value is ``30``. It must be non-negative.
 - ``retry_multiplier``: Factor to multiply the backoff after a failed
-  retry. Its default value is ``1``. It must be greater than or equal
+  retry. Its default value is ``1.05``. It must be greater than or equal
```
Kilo59, line 72: What happened to the old [`connection_attempt_limit `](http://hazelcast.github.io/hazelcast-python-client/3.10/hazelcast.config.html#hazelcast.config.ClientNetworkConfig.connection_attempt_limit) ?

Is there no longer an option to control the number of connection attempts?

mdumandag: Hi, this configuration option is removed from the client in the 4.0 release with a more enhanced exponential backoff feature. You can get the almost same behavior by configuring the `cluster_connect_timeout` parameter.

Kilo59: In my use case, I just want to either attempt 1 or 2 connection attempts and then stop (as part of an integration test).

The exponential backup is a nice feature (and it will be useful in production), but it looks like the logic for limiting the connection attempts is now somewhat convoluted.

mdumandag: I can agree on some of your points but let me clarify the `cluster_connect_timeout` a little bit. 

It is actually an absolute timeout for whole connection attempts.

Let say that you specified 3 possible member addresses in your configuration. Client first will store the time it started trying to connect the cluster. (Let say that it is `start_time`)

Client tries to connect one of the three. If it cannot connect any of them, then client tries to sleep a little bit. But before sleeping, it will check the the time passed since the first connection attempt(`time.time() - start_time`). If it is more than the `cluster_connect_timeout`, it will throw an exception saying that client could not connect to any of the members. If it is less than that, it will sleep and try to connect possible addresses again. This loops continues until the client reaches the `cluster_connect_timeout` or connects one of the members.

mdumandag: The same logic applies to reconnection process

Kilo59: @mdumandag thanks for the quick response and clarification, that's very helpful.

mdumandag: No problem at all, happy to help you. Let me rephrase that part tomorrow

docs/client_connection_strategy.rst
```diff
@@ -86,6 +87,7 @@ A pseudo-code is as follows:
     while (try_connect(connection_timeout)) != SUCCESS) {
         if (get_current_time() - begin_time >= CLUSTER_CONNECT_TIMEOUT) {
             // Give up to connecting to the current cluster and switch to another if exists.
+            // For the default values, CLUSTER_CONNECT_TIMEOUT is infinite.
```
yuce, line 90: How about: "CLUSTER_CONNECT_TIMEOUT is infinite by default."?

hazelcast/connection.py
```diff
@@ -280,11 +280,18 @@ def _trigger_cluster_reconnection(self):
             self._start_connect_to_cluster_thread()
 
     def _init_wait_strategy(self, config):
+        cluster_connect_timeout = config.cluster_connect_timeout
+        if cluster_connect_timeout == -1:
+            # If the no timeout is specified by the
+            # user, or set to -1 explicitly, set
+            # the timeout to infinite.
+            cluster_connect_timeout = six.MAXSIZE
```
yuce, line 288: As far as I've checked `sys.maxsize` is available on all versions of CPython. On Python 2.7 and Python 3.6 both evaluate `sys,maxsize` to `9223372036854775807`. So, I think it would be better to replace `sys.MAXSIZE` with `sys.maxsize`.

hazelcast/config.py
```diff
@@ -531,8 +531,8 @@ def __init__(self):
         self._retry_initial_backoff = 1.0
         self._retry_max_backoff = 30.0
         self._retry_jitter = 0.0
-        self._retry_multiplier = 1.0
-        self._cluster_connect_timeout = 120.0
+        self._retry_multiplier = 1.05
```
yuce, line 534: Can we move `1.05` to a global variable, something like `DEFAULT_RETRY_MULTIPLIER` and replace instances of it?

hazelcast/config.py
```diff
@@ -531,8 +531,8 @@ def __init__(self):
...
+        self._cluster_connect_timeout = -1
```
yuce, line 535: Can we move `-1` to a global variable, something like `DEFAULT_CLUSTER_CONNECT_TIMEOUT` and replace instances of it?

docs/client_connection_strategy.rst
```diff
@@ -69,13 +69,14 @@ The following are configuration element descriptions:
...
-  retry. Its default value is ``1``. It must be greater than or equal
+  retry. Its default value is ``1.05``. It must be greater than or equal
   to ``1``.
 - ``retry_jitter``: Specifies by how much to randomize backoffs. Its
   default value is ``0``. It must be in range ``0`` to ``1``.
 - ``cluster_connect_timeout``: Timeout value in seconds for the client
-  to give up to connect to the current cluster. Its default value is
-  ``120``.
+  to give up connecting to the cluster. Its default value is
+  ``-1``. For the default value, client will not stop trying to connect
```
yuce, line 78: I think we can rephrase "For the default value ..." as "By default ..."

yuce (changes requested):
The PR looks good other than a few nitpicks.

docs/client_connection_strategy.rst
yuce: Maybe even better: "The client will continuously try to connect by default."

yuce (approved):
LGTM

mdumandag:
Thanks for the review @yuce

PR #359, opened 2021-02-05
WIP config_dict

Kilo59:


hazelcast/config.py
```diff
@@ -1385,3 +1386,14 @@ def from_dict(cls, d):
...
                 )
         return config
+
+
+def config_dict(**kwargs) -> six.types.Dict[str, six.types.Any]:
+    """Create a valid dictionary of HazelcastClient configuration options."""
+    kwarg_keys = set(kwargs.keys())
+    valid_config_keys = {t[0] for t in inspect.getmembers(_Config, lambda o: isinstance(o, property))
+    if not kwarg_keys.subset(valid_config_keys):
+        raise TypeError("Invalid keyword arguments :{}".format(','.join(kwarg_keys.difference(valid_config_keys))))
```
Kilo59, line 1396: Maybe raise a deprecation warning here instead.

hazelcast/config.py
```diff
@@ -1385,3 +1386,14 @@ def from_dict(cls, d):
                     "Unrecognized config option for the flake id generator: %s" % k
                 )
         return config
+
+
+def config_dict(**kwargs) -> six.types.Dict[str, six.types.Any]:
+    """Create a valid dictionary of HazelcastClient configuration options."""
```
Kilo59, line 1392: This isn't mean to replace the current way of configuring the client, but to complement it.
```python
# both of these would still be valid

HazelcastClient(foo=bar, fizz=buzz, xml="my_settings.xml")

my_validated_config = config_dict(foo=bar,fizz=buzz, xml="my_settings.xml")
HazelcastClient(**my_validated_config)

```
The difference is I can be sure that my configuration settings are valid before I attempt to instantiate my client, and I can see how my configuration settings resolve when I'm combining XML and kwarg settings.

PR #365, opened 2021-02-09
Refactored configuration

yuce:
- Refactored _Config class.
- Made _Config experimentally public as Config.
- HazelcastClient accepts a Config object.

### DIFF
--- a/hazelcast/client.py
+++ b/hazelcast/client.py
@@ -1,14 +1,15 @@
 import logging
 import threading
+from typing import Optional
 
 from hazelcast import six
 from hazelcast.cluster import ClusterService, _InternalClusterService
-from hazelcast.config import _Config
+from hazelcast.config import _Config, Config
 from hazelcast.connection import ConnectionManager, DefaultAddressProvider
 from hazelcast.core import DistributedObjectInfo, DistributedObjectEvent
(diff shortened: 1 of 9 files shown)


tests/unittests/util_test.py
```diff
@@ -0,0 +1,49 @@
+import unittest
```
Kilo59, line 1: Have you guys considered switching to pytest?

hazelcast/client.py
```diff
@@ -50,6 +51,13 @@ class HazelcastClient(object):
     distributed data structures on the Hazelcast clusters.
 
     Keyword Args:
+        config (`hazelcast.config.Config`): Optional configuration object
```
Kilo59, line 54: Could `hazelcast.config.Config` inherit from `dict`?
or have a `to_dict()`/`from_dict()` method.

Then you could use dict unpacking to create the client with the config.
You could probably do this instead of accepting a `Config` object directly or make just another option.

Either way, it would make the transition from v3 easier.

```python
# lets log the config settings
LOGGER.info(my_config_obj.dict())

HazelcastClient(**my_config_obj.dict())
```

mdumandag:
closing this in favor of #521

PR #368, opened 2021-02-12
Do not overwrite invocation timeout in the constructor

mdumandag:
We were mistakenly overwriting the invocation timeout to None
even if it is explicitly set to something else.

We are now respecting the given timeout value in the invocation
constructor.


mdumandag:
@yuce can you also take a look at this PR?

yuce (approved):

PR #370, opened 2021-02-18
Use time.monotonic() instead of time.time() for profiling 

srknzl:
* [time.time()](https://docs.python.org/3/library/time.html#time.time) returns the time based on system clock, which can go back for several reasons such as 
    * synchronization from a web server
    * leap seconds
* Thus, using time.time() for profiling is not the correct option. The difference might even be negative due to reasons above.
* Instead of using time.time(), [time.monotonic()](https://docs.python.org/3/library/time.html#time.monotonic) should be used as it is based on cpu clock and always increases.

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -34,14 +34,14 @@ def fill_map(map, size=10, key_prefix="key", value_prefix="val"):
 
 
 def get_ssl_config(
-        cluster_name,
-        enable_ssl=False,
-        cafile=None,
-        certfile=None,
-        keyfile=None,
-        password=None,
-        protocol=SSLProtocol.TLSv1_2,
-        ciphers=None,
+    cluster_name,
+    enable_ssl=False,
+    cafile=None,
+    certfile=None,
+    keyfile=None,
+    password=None,
+    protocol=SSLProtocol.TLSv1_2,
+    ciphers=None,
 ):
     config = {
         "cluster_name": cluster_name,
--- a/tests/util.py
+++ b/tests/util.py
@@ -3,27 +3,12 @@ import time
 
 from uuid import uuid4
 from hazelcast.config import SSLProtocol
-from hazelcast.six import PY3
 
-
-def get_current_timestamp():
-    """
-    Get current timestamp.
-    time.monotonic() is more consistent since it uses cpu clock rather than system clock.
-
-    python versions 3.3 and up supports time.monotonic(). Before python version 3.5 and on gnu/hurd
-    monotonic() is not available. Since we support support python3.4 and up, there is a check
-    for the monotonic attribute availability, just for gnu/hurd.
-
-    Returns:
-        If time.time() used, this function returns is number of seconds since 1970,
-        if time.monotonic() is used, it still returns fractional seconds but its value alone
-        is not meaningful, but it can still can be used take time difference for profiling.
-    """
-    if PY3 and hasattr(time, "monotonic"):
-        return time.monotonic()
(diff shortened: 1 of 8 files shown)


mdumandag:
Hi @srknzl, Python client supports Python 2 and 3.4 which do not have the `monotonic` method defined. You need to use it conditionally depending on the Python version

srknzl:
Hi, @mdumandag. Done!

mdumandag:
verify

srknzl:
@mdumandag Actually, monotonic is added on version 3.3. We can use it for python3.4 too, but I am not sure what this means in the documentation 

`Changed in version 3.5: The function is now always available and always system-wide.`

PS: I used monotonic() on a python3.4 docker, and it works

mdumandag:
I was not sure either. https://docs.python.org/3.3/library/time.html#time.monotonic This states that it is available on lots of operating systems on 3.3. So, I guess since the scope of this change is tests only, and it seems to be available on lots of operating systems, we can use it in all Python3 versions.

BTW, you can use `PY3` constant defined in the `hazelcast/six.py` to check whether or not it is running on Python3

srknzl:
But not exactly for all python3 versions, at least we should check python version is more than 3.3. I used when version is >= 3.3. Can we verify again ?

mdumandag:
We don't support Python3 versions below 3.4, so it should be okay to just check it

srknzl:
Ok done

mdumandag:
verify

srknzl:
@mdumandag I found that on versions 3.3-3.5(not included) monotonic() might not be available on GNU/Hurd. Check https://www.python.org/dev/peps/pep-0418/#time-monotonic

mdumandag:
I don't think it will be a problem, but if you want to be sure and use it 3.4 version of most of the operating system, you could try to use logic like

```python
if hasattr(time, "monotonic"):
    get_current_timestamp = time.monotonic
else:
    get_current_timestamp = time.time
```

in tests.util

srknzl:
Ok, if the change is fine, let's verify again

srknzl:
Sorry let me change the comment

srknzl:
Done.

tests/util.py
```diff
@@ -3,6 +3,27 @@
 
 from uuid import uuid4
 from hazelcast.config import SSLProtocol
+from hazelcast.six import PY3
+
+
+def get_current_timestamp():
```
mdumandag, line 9: Instead of defining this as a function, and checking whether or not `time` has `monotonic` attribute each time this is called, we can define this function like
```python
if hasattr(time, "monotonic"):
    get_current_timestamp = time.monotonic
else:
    get_current_timestamp = time.time
```

Note that, we don't need to check for PY3 anymore, we will simply use it if we have it

srknzl: Ok check it again pls

mdumandag:
verify

srknzl:
Why python3 test build failed

mdumandag:
@srknzl Checkstyle failed. Please use black to format the code as described in https://github.com/hazelcast/hazelcast-python-client#development

mdumandag:
verify

srknzl:
Yeah, I did not see it sorry. Will keep it in mind next time. Plus, I ran the tests and they pass.

mdumandag (approved):

mdumandag:
Thanks for the contribution @srknzl, I am merging the PR

PR #380, opened 2021-03-10
Fail fast on the METADATA group name for CP data structures

### LINKED ISSUE #374: Do not allow using METADATA as the group name on the CP data structures
`METADATA` is a special group name dedicated to the internal CP group that is used for maintaining CP groups and members. As of now, when you try to create a CP data structure with the `METADATA` as the group name, you receive an error from the server-side. However, it would be good if we can fail fast on the client-side.

So, a code like this should fail on the client-side.

```python
import hazelcast

client = hazelcast.HazelcastClient()

view_counter = client.cp_subsystem.get_atomic_long("views@METADATA")
```

mdumandag:
It is not possible to run the CP data structures on the metadata
group name. We now fail-fast if someone tries to use the `METADATA`
as the CP group name. Also, the group name comparison made
case-insensitive.

Closes #374

### DIFF
--- a/hazelcast/cp.py
+++ b/hazelcast/cp.py
@@ -155,6 +155,7 @@ class CPSubsystem(object):
 
 
 _DEFAULT_GROUP_NAME = "default"
+_METADATA_CP_GROUP_NAME = "metadata"
 
 
 def _without_default_group_name(name):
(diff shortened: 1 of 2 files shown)


srknzl (approved):
Looking good. I wonder why metadata cannot be used as group name

mdumandag:
See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/cp/CPGroup.java#L37, it is used by Hazelcast internally

PR #381, opened 2021-03-10
Add tests that verify near cache works even if the cluster connecti…

### LINKED ISSUE #309: [TRACKING ISSUE] add test for ClientNonStopNearCache scenario

**add test for ClientNonStopNearCache scenario**

See https://github.com/hazelcast/hazelcast/pull/16717 for details.

---

mdumandag:
…on is down

Similar to the Java client, added simple tests that verify the map with
near cache continues to work for the entries that exist in the cache
even if the cluster connection is down.

Closes #309

### DIFF
--- a/hazelcast/near_cache.py
+++ b/hazelcast/near_cache.py
@@ -84,6 +84,7 @@ class NearCache(dict):
         eviction_sampling_count=None,
         eviction_sampling_pool_size=None,
     ):
+        super(NearCache, self).__init__()
         self.name = name
         self.serialization_service = serialization_service
         self.in_memory_format = in_memory_format
--- a/tests/integration/backward_compatible/client_test.py
+++ b/tests/integration/backward_compatible/client_test.py
@@ -33,7 +33,6 @@ class ClientTest(HazelcastTestCase):
             events = []
 
             def event_collector(e):
-                print(e)
                 if e == LifecycleState.DISCONNECTED:
                     events.append(e)
 
--- a/tests/integration/backward_compatible/proxy/map_nearcache_test.py
+++ b/tests/integration/backward_compatible/proxy/map_nearcache_test.py
@@ -1,11 +1,14 @@
 import os
 
+from hazelcast.config import ReconnectMode
+from hazelcast.errors import ClientOfflineError
+from hazelcast.lifecycle import LifecycleState
 from tests.hzrc.ttypes import Lang
 
-from tests.base import SingleMemberTestCase
+from tests.base import SingleMemberTestCase, HazelcastTestCase
 from tests.util import random_string
 from hazelcast.six.moves import range
-from hazelcast import six
+from hazelcast import six, HazelcastClient
 
 
 class MapTest(SingleMemberTestCase):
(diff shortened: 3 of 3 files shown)


hazelcast/near_cache.py
```diff
@@ -84,6 +84,7 @@ def __init__(
         eviction_sampling_count=None,
         eviction_sampling_pool_size=None,
     ):
+        super(NearCache, self).__init__()
```
srknzl, line 87: Why you added this? I guess you forget to add this before

srknzl: BTW, we may not need extending from a dict. Where do we need dict extension? I think we can only implement __getitem__ and __setitem__ to getter and setter behaviour with [].

tests/integration/backward_compatible/client_test.py
```diff
@@ -33,7 +33,6 @@ def lifecycle_event_collector():
             events = []
 
             def event_collector(e):
-                print(e)
```
srknzl, line 36: I guess this was for debugging and forgotten

tests/integration/backward_compatible/proxy/map_nearcache_test.py
```diff
@@ -108,3 +111,71 @@ def fill_map_and_near_cache(self, count=10):
...
+        collector = event_collector()
+
+        client = HazelcastClient(
+            cluster_name=cluster.id,
+            reconnect_mode=ReconnectMode.ASYNC,
+            near_caches={"map": {}},
+            lifecycle_listeners=[collector],
+        )
+
+        map = client.get_map("map").blocking()
```
srknzl, line 144: This blocking/non-blocking does not exist in Java client right? or in java the default is blocking ?

tests/integration/backward_compatible/proxy/map_nearcache_test.py
```diff
@@ -108,3 +111,71 @@ def fill_map_and_near_cache(self, count=10):
...
+            map.put(i, i)
+
+        # Populate the near cache
+        for i in range(ENTRY_COUNT):
+            map.get(i)
+
+        rc.terminateCluster(cluster.id)
+        rc.exit()
+
+        self.assertTrueEventually(lambda: self.assertEqual(1, len(collector.events)))
```
srknzl, line 156: Is there any possibility for the disconnected event to fire more than once? If so, the test will be flaky

tests/integration/backward_compatible/proxy/map_nearcache_test.py
```diff
@@ -108,3 +111,71 @@ def fill_map_and_near_cache(self, count=10):
...
+
+    def test_put_to_map_when_the_cluster_is_down(self):
+        with self.assertRaises(ClientOfflineError):
+            self.map.put(ENTRY_COUNT, ENTRY_COUNT)
+
+    @staticmethod
+    def read_cluster_config():
+        path = os.path.abspath(__file__)
+        dir_path = os.path.dirname(path)
+        with open(os.path.join(dir_path, "hazelcast.xml")) as f:
```
srknzl, line 180: Asking just because I wonder. What is this hazelcast.xml? What kind of config it stores? And, why we don't need to create a cluster in the test in java side?(Only a single member created without a special config as far as I see )

hazelcast/near_cache.py
mdumandag: 1. Yes
2. It was like this before and I don't see a point in changing it.

tests/integration/backward_compatible/client_test.py
mdumandag: Yes

tests/integration/backward_compatible/proxy/map_nearcache_test.py
mdumandag: Yes, but it doesn't matter

tests/integration/backward_compatible/proxy/map_nearcache_test.py
mdumandag: Nope, the client will disconnect only once

tests/integration/backward_compatible/proxy/map_nearcache_test.py
mdumandag: This is the server-side config we generally use in tests. Check the file in this folder for its content. It makes the test more resilient to batch invalidations. Note that, Python side of the Near Cache is not as complete as the Java side so a special config is needed here.

tests/integration/backward_compatible/proxy/map_nearcache_test.py
srknzl: Yes to which question 😄  I guess to second one. I asked this just to learn as well. We have getAsync etc in the java client. I guess they are the non-blocking ones.

srknzl (approved):
LGTM 👍🏻

tests/integration/backward_compatible/proxy/map_nearcache_test.py
mdumandag: The second, it exists for some methods as a `...Async` variant

PR #383, opened 2021-03-12
Separate the backward compatible integration tests

mdumandag:
Marked all the integration tests as backward compatible after
verifying that they run just fine against the v4.0 client.

To make it happen, some missing functionalities are conditionally
used via `backward_compatible/util.py`. In some tests, we might be
using private APIs but I have verified that they also exist in the
older versions. In the future, if we want to change private API,
we can either conditionally use them or mark some tests as
backward-incompatible by moving them out of this folder.


yuce (approved):
Thanks, LGTM!

PR #385, opened 2021-03-23
Do not use deprecated abstract classes from collections

### LINKED ISSUE #384: Deprecation warning: util.py:4
I'm getting this warning:
```
$ import hazelcast
/home/jara/.local/lib/python3.8/site-packages/hazelcast/util.py:4: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import Sequence, Iterable
```
I believe I have the most up to date HZ client version, but it's possible I am just clumsy with pip :)

mdumandag:
Abstract classes Sequence and Iterable are meant to be imported
from `collections.abc` instead of `collections` starting from
Python 3.3. Since the minimum supported Python version for the
client is 3.4, we can safely try to import it from the new location
and fallback to the old location if the import fails. It will only
fail for Python 2.7 in which it is fine to import those abstract
classes from `collections`.

Closes #384

### DIFF
--- a/hazelcast/util.py
+++ b/hazelcast/util.py
@@ -1,7 +1,12 @@
 import random
 import threading
 import time
-from collections import Sequence, Iterable
+
+try:
+    from collections.abc import Sequence, Iterable
+except ImportError:
+    # Compatibility for Python2
+    from collections import Sequence, Iterable
 
 from hazelcast import six
 


srknzl (approved):
Can you reproduce the warning in the issue? I could not do it.

mdumandag:
Strangely, I also cannot reproduce the warning (tested on python 3.6, 3.7, and 3.9) and our nightly python 3.9 test runner works just fine but I see that it is merged in https://github.com/python/cpython/commit/ef092fe9905f61ca27889092ca1248a11aa74498

Anyways, let's use this version just to be sure

PR #386, opened 2021-03-23
Skip tests when a newer server is expected

mdumandag:
On certain tests, a newer server might be required to test functionality.
Formerly, we were using attributes to distinguish such tests but this
way, it is easier to handle all cases.

Also, removed some unneeded compatibility code.

### DIFF
--- a/tests/integration/backward_compatible/proxy/cp/atomic_long_test.py
+++ b/tests/integration/backward_compatible/proxy/cp/atomic_long_test.py
@@ -1,7 +1,7 @@
 from hazelcast.errors import DistributedObjectDestroyedError
 from hazelcast.serialization.api import IdentifiedDataSerializable
 from tests.integration.backward_compatible.proxy.cp import CPTestCase
-from tests.util import set_attr
+from tests.util import mark_server_version_at_least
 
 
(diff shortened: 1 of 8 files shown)


tests/integration/backward_compatible/proxy/cp/atomic_long_test.py
```diff
@@ -106,30 +106,30 @@ def test_set(self):
         self.assertIsNone(self.atomic_long.set(42))
         self.assertEqual(42, self.atomic_long.get())
 
-    @set_attr(min_server_version="4.1")
```
srknzl, line 109: Was the backward compatibility test runner supposed to check this attribute?

srknzl (approved):

tests/integration/backward_compatible/proxy/cp/atomic_long_test.py
mdumandag: I changed the runner so that it doesn't check it anymore. We dynamically skip tests at runtime now, just like the Node.js client

srknzl: Just like I predicted, good 👍🏻

PR #387, opened 2021-04-02
Do not allow using duplicate field names in ClassDefinitionBuilder

### LINKED ISSUE #330: [TRACKING ISSUE] Check duplicate fields when building class definition

**Check duplicate fields when building class definition**

See https://github.com/hazelcast/hazelcast/pull/17682 for details.

---

ClassDefinitionBuider.addX methods checks if a field
is already added with same name and
throws HazelcastSerializatioException if already added.

fixes https://github.com/hazelcast/hazelcast/issues/17616

mdumandag:
With this change, `add_xxx_field` methods of the ClassDefinitionBuilder
checks the class name first, and if it already registered, raises
HazelcastSerializationError.

Also, missing docstrings for the `add_xxx_field` methods are added.

Closes #330

### DIFF
--- a/hazelcast/serialization/portable/classdef.py
+++ b/hazelcast/serialization/portable/classdef.py
@@ -132,6 +132,16 @@ class ClassDefinition(object):
 
 
 class ClassDefinitionBuilder(object):
+    """Builder class to construct :class:`ClassDefinition` of
+    :class:`hazelcast.serialization.api.Portable` objects.
+
+    One must make sure that the order of the fields added to this
+    builder is consistent across all the usages of the Portable
+    object such as the write order of the fields of the
+    :func:`Portable.write_portable <hazelcast.serialization.api.Portable.write_portable>`
+    method.
+    """
+
     def __init__(self, factory_id, class_id, version=0):
         self.factory_id = factory_id
         self.class_id = class_id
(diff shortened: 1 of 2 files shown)


yemreinci (approved):
Looks OK.

hazelcast/serialization/portable/classdef.py
```diff
@@ -210,71 +382,219 @@ def add_portable_array_field(self, field_name, class_def):
...
+                after :func:`build` or a field with the same
+                name is already registered.
+
         .. deprecated:: 4.1
             This method is deprecated and will be removed in the
             next major version. Use :func:`add_string_array_field` instead.
         """
         return self.add_string_array_field(field_name)
 
     def add_field_def(self, field_def):
```
sancar, line 567: I think this is public by accident as it is in java. Can we add documentation to tell the users that this is private(I am assuming we can not delete it anymore)

Can we add a documentation about the order that fields added should be consistent across all languages and also it should be consistent with write order in `writePortable` on ClassDefinitionBuilder ?

mdumandag: Added a warning to this method and a note about the field order in the ClassDefinitionBuilder docstring

sancar (approved):

PR #390, opened 2021-04-15
[API-282] Implement SQL service

mdumandag:
Added a new service called `sql` that executes SQL queries on
Maps.

The service returns a `SqlResult` immediately so that, the user
can cancel queries immediately.

Also, via `SqlResult`, the user can get blocking or non-blocking
iterators to get results row-by-row.

The rows contain row metadata along with the objects that can be
accessed either through the column names or column indexes.

The row metadata contains information about the columns returned
in rows.

Also, the users can get the update count, row metadata, and whether or
not the result contains rows through `SqlResult`.


mdumandag:
verify

mdumandag:
verify

tests/integration/backward_compatible/sql_test.py
```diff
@@ -49,6 +59,9 @@ def _populate_map(self, entry_count=10, value_factory=lambda v: v):
         self.map.put_all(entries)
 
 
+@unittest.skipIf(
+    is_client_version_older_than("4.2"), "Tests the features added in 4.2 version of the client"
+)
```
srknzl, line 64: Can we add server version >= 4.2 as well?

srknzl: Sorry I saw it later on

mdumandag:
verify

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+      conflict.
+    - For :class:`Portable <hazelcast.serialization.api.Portable>` objects,
+      field names used in the
+      :func:`write_portable() <hazelcast.serialization.api.Portable.write_portable>`
+      method are used to populate the column list.
+
+    The whole key and value objects could be accessed through special fields
+    ``__key`` and ``this``, respectively. If key (or value) object has fields,
+    then the whole key (or value) field is exposed as a normal field. Otherwise the
+    field is hidden. Hidden fields can be accessed directly, but are not returned
```
srknzl, line 65: I think the correct sentence is(as I told you in a message I sent to you): 

If key (or value) object has fields, then the key (or value) field is hidden. Otherwise, the whole key (or value) field is exposed as a normal field.

tests/unit/sql_test.py
```diff
@@ -297,6 +296,47 @@ def test_statement_expected_result_type(self):
...
+
+        invalid_inputs = [4, 5, "6", None]
+        for invalid in invalid_inputs:
+            with self.assertRaises((IndexError, AssertionError)):
+                row_metadata.get_column(invalid)
+
+    def test_row_metadata_find_column(self):
+        row_metadata = self._create_row_metadata()
+
+        valid_inputs = ["0", "1", "2", "-1"]
```
srknzl, line 323: some clever way of using -1 here :)

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+            sql (str): SQL string.
+            *params: Query parameters that will be passed to
+                :func:`SqlStatement.add_parameter`.
+
+        Returns:
+            SqlResult: The execution result.
+
+        Raises:
+            HazelcastSqlError: In case of execution error.
+            AssertionError: If the SQL parameter is not a string.
```
srknzl, line 161: ValueError: If the SQL string is empty

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+
+    def iterator(self):
+        """Returns the iterator over the result rows.
+
+        The iterator may be requested only once.
+
+        Returns:
+            Future[Iterator[Future[SqlRow]]]: Iterator that produces Future
+            of :class:`SqlRow` s. See the class documentation for the correct
+            way to use this.
```
srknzl, line 914: Raises:

HazelcastSqlError: In case of an SQL execution error.
ValueError: If the result only contains update count, or the iterator is already requested.

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+        """
+        return self._get_iterator(False)
+
+    def is_row_set(self):
+        """
+        Returns:
+            Future[bool]: Whether this result has rows to iterate.
```
srknzl, line 921: Raises:

HazelcastSqlError: In case of an SQL execution error.

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+            # we only got the update count.
+            return response.row_metadata is not None
+
+        return self._execute_response.continue_with(continuation)
+
+    def update_count(self):
+        """Returns the number of rows updated by the statement or ``-1`` if this
+        result is a row set. In case the result doesn't contain rows but the
+        update count isn't applicable or known, ``0`` is returned.
+
```
srknzl, line 936: Raises:

HazelcastSqlError: In case of an SQL execution error.

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+            response = future.result()
+            # This will be set to -1, when we got row set on the client side.
+            # See _on_execute_response.
+            return response.update_count
+
+        return self._execute_response.continue_with(continuation)
+
+    def get_row_metadata(self):
+        """Gets the row metadata.
+
```
srknzl, line 951: Raises:

HazelcastSqlError: In case of an SQL execution error.
ValueError: If the result contains only update count.

(Note: raises may not be correct since future is rejected with these errors, you can find the correct wording)

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+    def close(self):
+        """Release the resources associated with the query result.
+
+        The query engine delivers the rows asynchronously. The query may
+        become inactive even before all rows are consumed. The invocation
+        of this command will cancel the execution of the query on all members
+        if the query is still active. Otherwise it is no-op. For a result
+        with an update count it is always no-op.
+
```
srknzl, line 974: Raises:

HazelcastSqlError: In case of an SQL error.

hazelcast/sql.py
```diff
@@ -0,0 +1,1629 @@
...
+        self._schema = None
+        self._expected_result_type = SqlExpectedResultType.ANY
+
+    @property
+    def sql(self):
+        """str: The SQL string to be executed."""
+        return self._sql
+
+    @sql.setter
+    def sql(self, sql):
```
srknzl, line 1467: If we can add api doc for setters I would rather add them. Please add "Raises:" too

srknzl (approved):
Approving with minor comments. Some of them are in this pr some of them are in https://github.com/hazelcast/hazelcast-python-client/pull/406

PR #391, opened 2021-04-19
Use assertTrueEventually in distributed object tests

mdumandag:
Improved the distributed object tests slightly by using assertTrueEventually.

Although we have never seen it before, this approach is safer.


srknzl (approved):
👍🏻

Issue #378, opened 2021-03-10
tests.proxy.cp.fenced_lock_test.FencedLockTest.test_lock_after_client_shutdown


mdumandag:
**Error Message**
```Exception must be of BaseException type```

**Stacktrace**

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/jenkins/jenkins_slave/workspace/python2.7-lab-master/tests/proxy/cp/fenced_lock_test.py", line 50, in test_lock_after_client_shutdown
    another_client.shutdown()
  File "/home/jenkins/jenkins_slave/workspace/python2.7-lab-master/hazelcast/client.py", line 647, in shutdown
    self._connection_manager.shutdown()
  File "/home/jenkins/jenkins_slave/workspace/python2.7-lab-master/hazelcast/connection.py", line 199, in shutdown
    connection.close("Hazelcast client is shutting down", None)
  File "/home/jenkins/jenkins_slave/workspace/python2.7-lab-master/hazelcast/connection.py", line 773, in close
    self._connection_manager.on_connection_close(self, cause)
  File "/home/jenkins/jenkins_slave/workspace/python2.7-lab-master/hazelcast/connection.py", line 235, in on_connection_close
    pending.set_exception(cause)
  File "/home/jenkins/jenkins_slave/workspace/python2.7-lab-master/hazelcast/future.py", line 45, in set_exception
    raise RuntimeError("Exception must be of BaseException type")
RuntimeError: Exception must be of BaseException type
```

PR #392, opened 2021-04-20
[API-337] Monkey patch client shutdown on tests for older client versions

mdumandag:
Due to a bug on the older client releases, the backward compatibility
tests sometimes fail on the client.shutdown() call. The reasoning
is explained in the https://github.com/hazelcast/hazelcast-python-client/issues/378
and linked PRs.

To solve this problem, we monkey patch the shutdown method, try to
call it, if it fails with an expected exception, cleanup the
leftovers and return without an error.

Shutdown calls generally happen on the teardown methods of the
tests so that we do not leak any clients to other tests. This
change should not cause any problems in actual tests.

This change should not cause any change in the current codebase
since we increased the client version to the next version in this PR (to 4.1)

### DIFF
--- a/hazelcast/__init__.py
+++ b/hazelcast/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "4.0"
+__version__ = "4.1"
 
 # Set the default handler to "hazelcast" loggers
 # to avoid "No handlers could be found" warnings.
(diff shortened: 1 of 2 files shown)


srknzl (approved):
lgtm

PR #394, opened 2021-04-21
Properly shutdown clients in reconnect tests

mdumandag:
Client shutdown calls were missing for some of the reconnect tests.
Added that and added a secondary shutdown mechanism in case something
goes wrong in the tests (like assertion failures) and we couldn't shutdown
the client.

### DIFF
--- a/tests/integration/backward_compatible/reconnect_test.py
+++ b/tests/integration/backward_compatible/reconnect_test.py
@@ -165,8 +165,16 @@ class ReconnectWithDifferentInterfacesTest(HazelcastTestCase):
 
     def setUp(self):
         self.rc = self.create_rc()
+        self.client = None
 
     def tearDown(self):
+        if self.client:
+            # If the test is failed, and we couldn't shutdown
+            # the client, try to shutdown here to make sure that
+            # we are not going to affect other tests. If the client
+            # is already shutdown, then this is basically no-op.
+            self.client.shutdown()
+
         self.rc.exit()
 
     def test_connection_count_after_reconnect_with_member_hostname_client_ip(self):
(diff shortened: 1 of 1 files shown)


tests/integration/backward_compatible/reconnect_test.py
```diff
@@ -268,3 +280,6 @@ def assertion():
             self.assertNotEqual(0, event_count.get())
 
         self.assertTrueEventually(assertion)
+
+        client.shutdown()
+        self.rc.terminateCluster(cluster.id)
```
srknzl, line 285: how was the client shutting down before this added?

tests/integration/backward_compatible/reconnect_test.py
```diff
@@ -244,6 +254,8 @@ def _verify_listeners_after_client_disconnected(self, member_address, client_add
             cluster_connect_timeout=sys.maxsize,
         )
 
+        self.client = client
+
```
srknzl, line 258: nit: could do this as well (same for other `self.client = client`):
```python
self.client = HazelcastClient(
            cluster_name=cluster.id,
            cluster_members=[client_address],
            cluster_connect_timeout=sys.maxsize,
 )
```

srknzl (approved):
LGTM 👍🏻  Left a question and a nit

tests/integration/backward_compatible/reconnect_test.py
mdumandag: It was not :D That was the problem

srknzl: ok that question was to clarify 😄 , thanks

tests/integration/backward_compatible/reconnect_test.py
mdumandag: Then I head to reference them as self.client instead of client, which I don't like

mdumandag:
Thanks for the review Serkan

PR #395, opened 2021-04-26
[API-339] Implement ReliableTopic

mdumandag:
Implemented ReliableTopic on top of the Ringbuffer. Like the
interface provided in the Java-side, we are offering methods
to add/remove listeners and publish messages in batches or
one by one. The provided API is non-blocking just like the
other parts of the client.

Documentation and code samples will be sent in a separate PR.

Thanks a lot to @buraksezer for the initial version of this PR
which we used as a base for this feature.


mdumandag:
Failed since the SSL certificates we used are expired now. Will send a separate PR for them once the Hazelcast Enterprise side is merged and then I will restart these tests

tests/integration/backward_compatible/proxy/reliable_topic_test.py
```diff
@@ -0,0 +1,448 @@
...
+
+        class Listener(ReliableMessageListener):
+            def on_message(self, message):
+                pass
+
+            def retrieve_initial_sequence(self):
+                return -1
+
+            def store_sequence(self, sequence):
+                sequences.append(sequence)
```
sancar, line 140: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/topic/impl/reliable/DurableSubscriptionTest.java

Can you add a test like this to see store and retrieve is working correctly.

mdumandag: Added the same test

sancar (approved):

PR #405, opened 2021-05-06
Pretty print cluster connect timeout in logs

### LINKED ISSUE #404: [TRACKING ISSUE] [PLAT-305] Pretty print cluster connect timeout in logs

See https://github.com/hazelcast/hazelcast/pull/18642 for details.

---

The default value for the cluster connect timeout is infinite,
so it is represented with `Long.MAX_VALUE`. When we try to log
this as it is, we output something like below.

```
... cluster connect timeout: 9223372036854775807 ms, ...
```

To make this more clear, we will replace this with the
output below.

```
... cluster connect timeout: INFINITE, ...
```

If the user has provided some other timeout value (such as 999 ms),
the output will still be as it is now.

```
... cluster connect timeout: 999 ms, ...
```

mdumandag:
Before this change, we were logging the cluster connect timeout
as it is. Its default value is equal to `sys.maxsize`, so the
logs were like

```
... cluster connect timeout: 92233720368547758067s ...
```

We now output `INFINITE` for the default value.

```
... cluster connect timeout: INFINITE
```

If the user provides a different value, it will be printed
as it is now.

Closes #404

### DIFF
--- a/hazelcast/connection.py
+++ b/hazelcast/connection.py
@@ -2,7 +2,6 @@ import io
 import logging
 import random
 import struct
-import sys
 import threading
 import time
 import uuid
@@ -33,6 +32,8 @@ from hazelcast.util import AtomicInteger, calculate_version, UNKNOWN_VERSION
 
 _logger = logging.getLogger(__name__)
 
+_INF = float("inf")
+
 
 class _WaitStrategy(object):
     def __init__(self, initial_backoff, max_backoff, multiplier, cluster_connect_timeout, jitter):
@@ -45,7 +46,7 @@ class _WaitStrategy(object):
         self._cluster_connect_attempt_begin = None
         self._current_backoff = None
 
-        if cluster_connect_timeout == sys.maxsize:
+        if cluster_connect_timeout == _INF:
             self._cluster_connect_timeout_text = "INFINITE"
         else:
             self._cluster_connect_timeout_text = "%ds" % self._cluster_connect_timeout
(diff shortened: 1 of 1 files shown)


hazelcast/connection.py
```diff
@@ -45,6 +45,11 @@ def __init__(self, initial_backoff, max_backoff, multiplier, cluster_connect_tim
         self._cluster_connect_attempt_begin = None
         self._current_backoff = None
 
+        if cluster_connect_timeout == sys.maxsize:
```
yuce, line 48: `sys.maxsize` is an int, as far as I see we document `cluster_connect_timeout` as a float. Would using `float('inf')` work better?

yuce (changes requested):

hazelcast/connection.py
mdumandag: Changed it as you suggested

ihsandemir (approved):

hazelcast/connection.py
```diff
@@ -45,6 +46,11 @@ def __init__(self, initial_backoff, max_backoff, multiplier, cluster_connect_tim
         self._cluster_connect_attempt_begin = None
         self._current_backoff = None
 
+        if cluster_connect_timeout == _INF:
+            self._cluster_connect_timeout_text = "INFINITE"
+        else:
+            self._cluster_connect_timeout_text = "%ds" % self._cluster_connect_timeout
```
yuce, line 52: Should we format the timeout using something like `%.2f` in case someone uses a fractional timeout?

yuce (changes requested):
Thanks

yuce (approved):
Thanks, LGTM

PR #406, opened 2021-05-10
[API-306] Add SQL code sample 

mdumandag:
Added a code sample that demonstrates different ways to interact
with the SQL service and the returned SqlResult objects.

Also, removed the explanation of the folders in the README because
it was out of sync with the current folder structure and was
not offering lots of value. It was mainly just repeating the
folder names.

### DIFF
--- a/examples/sql/sql_example.py
+++ b/examples/sql/sql_example.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import hazelcast
 from hazelcast.serialization.api import Portable
 
--- a/examples/sql/sql_example.py
+++ b/examples/sql/sql_example.py
@@ -2,7 +2,6 @@ from __future__ import print_function
 
 import hazelcast
 from hazelcast.serialization.api import Portable
-
 from hazelcast.sql import SqlStatement
 
 
@@ -46,7 +45,6 @@ for row in result:
     name = row.get_object("name")
     print(name)
 
-
 # One can select all fields with *.
 # Also, with statement can be used to close the resources on the
 # server-side if something goes bad while iterating over rows.
@@ -67,7 +65,6 @@ with client.sql.execute("SELECT * FROM customers") as result:
         is_active = row.get_object_with_index(is_active_index)
 
         print(name, age, is_active)
-
 
 # Construct a statement object to control the properties of the query
 # Special keywords __key and this can be used to refer to key and value.
--- a/examples/sql/sql_example.py
+++ b/examples/sql/sql_example.py
@@ -27,6 +27,9 @@ class Customer(Portable):
     def get_class_id(self):
         return 1
 
+    def __repr__(self):
+        return "Customer(name=%s, age=%s, is_active=%s)" % (self.name, self.age, self.is_active)
+
 
 client = hazelcast.HazelcastClient(portable_factories={1: {1: Customer}})
 
@@ -85,8 +88,8 @@ with client.sql.execute_statement(statement) as result:
 
         print(key, age)
 
-# Parameters can also be specified for the basic query execution
-result = client.sql.execute("SELECT this FROM customers WHERE age > ?", 40)
+# Parameters can be passed directly in the basic execution syntax
+result = client.sql.execute("SELECT this FROM customers WHERE age > ? AND age < ?", 30, 40)
 
 for row in result:
     customer = row.get_object("this")
--- a/examples/sql/sql_example.py
+++ b/examples/sql/sql_example.py
@@ -1,0 +1,99 @@
+import hazelcast
+from hazelcast.serialization.api import Portable
+
+from hazelcast.sql import SqlStatement
+
+
+class Customer(Portable):
+    def __init__(self, name=None, age=None, is_active=None):
+        self.name = name
+        self.age = age
+        self.is_active = is_active
+
+    def write_portable(self, writer):
+        writer.write_string("name", self.name)
+        writer.write_int("age", self.age)
+        writer.write_boolean("is_active", self.is_active)
+
+    def read_portable(self, reader):
+        self.name = reader.read_string("name")
+        self.age = reader.read_int("age")
+        self.is_active = reader.read_boolean("is_active")
+
+    def get_factory_id(self):
+        return 1
+
+    def get_class_id(self):
+        return 1
+
+
+client = hazelcast.HazelcastClient(portable_factories={1: {1: Customer}})
+
+customers = client.get_map("customers").blocking()
+
+# Fill the customers map with some values
+customers.set(1, Customer("Peter", 42, True))
+customers.set(2, Customer("John", 23, False))
+customers.set(3, Customer("Joe", 33, True))
+
+# Project a single column that fits the criterion
+result = client.sql.execute("SELECT name FROM customers WHERE age < 35 AND is_active")
+
+for row in result:
+    # Get the object with the given column name in the row
+    name = row.get_object("name")
+    print(name)
+
+
+# One can also select all fields
+
+# Use with statement to close the resources on the server-side if something
+# goes bad while iterating over rows.
+
+with client.sql.execute("SELECT * FROM customers") as result:
+    for row in result:
+        # Get the objects with column names
+        name = row.get_object("name")
+        age = row.get_object("age")
+
+        # Get the metadata associated with the row
+        row_metadata = row.metadata
+
+        # Get the index of the is_active column
+        is_active_index = row_metadata.find_column("is_active")
+
+        # Get the object with the column index
+        is_active = row.get_object_with_index(is_active_index)
+
+        print(name, age, is_active)
+
+
+# Construct a statement object to control the properties of the query
+
+# Special keywords __key and this can be used to refer to key and value.
+# Also, a placeholder parameters can be specified
+statement = SqlStatement("SELECT __key, age FROM customers WHERE name LIKE ?")
+
+# Parameters will replace the placeholders on the server side
+statement.add_parameter("Jo%")
+statement.timeout = 5
+
+with client.sql.execute_statement(statement) as result:
+    # Row metadata can also be retrieved from the result
+    row_metadata = result.get_row_metadata().result()
+
+    for row in result:
+        key = row.get_object("__key")
+        age = row.get_object("age")
+
+        print(key, age)
+
+# Parameters can also be specified for the basic query execution
+result = client.sql.execute("SELECT this FROM customers WHERE age > ?", 40)
+
+for row in result:
+    customer = row.get_object("this")
+    print(customer)
+
+# Query can be closed explicitly
+result.close().result()
--- a/examples/sql/sql_example.py
+++ b/examples/sql/sql_example.py
@@ -47,10 +47,9 @@ for row in result:
     print(name)
 
 
-# One can also select all fields
-
-# Use with statement to close the resources on the server-side if something
-# goes bad while iterating over rows.
+# One can select all fields with *.
+# Also, with statement can be used to close the resources on the
+# server-side if something goes bad while iterating over rows.
 
 with client.sql.execute("SELECT * FROM customers") as result:
     for row in result:
@@ -71,7 +70,6 @@ with client.sql.execute("SELECT * FROM customers") as result:
 
 
 # Construct a statement object to control the properties of the query
-
 # Special keywords __key and this can be used to refer to key and value.
 # Also, a placeholder parameters can be specified
 statement = SqlStatement("SELECT __key, age FROM customers WHERE name LIKE ?")
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,28 +1,22 @@
  Hazelcast Python Client Code Examples Readme
 =============================
 
-This folder contains a collection of code samples which you can use to learn how to use Hazelcast features.
+This folder contains a collection of code samples which you can 
+use to learn how to use Hazelcast features.
 
 How to try these examples
 -------------------------
 
-* To try these examples, you should have a running Hazelcast member. If you are already familiar with Hazelcast and have a member running, add its address and port to your configuration as shown in the [**/learning-basics/1-configure_client.py**](learning-basics/1-configure_client.py).
-* If not, follow the steps described in the [Working with Hazelcast IMDG Clusters section](https://github.com/hazelcast/hazelcast-python-client/blob/master/README.md#12-working-with-hazelcast-imdg-clusters) of the README and start a Hazelcast member in your local computer.
-    
-Included example folders
------------------------
+* To try these examples, you should have a running Hazelcast member. 
+  If you are already familiar with Hazelcast and have a member running, 
+  add its address and port to your configuration as shown in the 
+  [**/learning-basics/1-configure_client.py**](learning-basics/1-configure_client.py).
+* If not, you can use our official Docker images to start a member.
 
-*   **/cloud-discovery** — Includes a code sample that shows how to use Hazelcast.Cloud
-*   **/learning-basics** — Code samples to show some Hazelcast basics like creating, configuring and destroying Hazelcast instances, configuring logging and Hazelcast configuration.
-*   **/list** — Contains a code sample to show usage of distributed list.
-*   **/map** — Code samples folder that includes some features of the Hazelcast distributed map in action.
-*   **/monitoring** — Includes code samples that show how to check status of Hazelcast instances using listeners.
-*   **/multimap** — Contains a code sample to show usage of distributed multi map.
-*   **/org-website** — Contains code samples that are shown on the https://hazelcast.org.
-*   **/queue** — Includes a code sample to show usage of distributed queue.
-*   **/ring-buffer** — Contains a code sample to show usage of distributed ring buffer.
-*   **/serialization** — Includes code samples that implement various serialization interfaces like IdentifiedDataSerializable, Portable. It also has code samples to show how to plug a custom and global serializers using StreamSerializer.
-*   **/set** — Contains a code sample to show usage of distributed set.
-*   **/ssl** — Contains code samples that show how to configure client with SSL or SSL with mutual authentication
-*   **/topic** — Includes a code sample to show usage of distributed topic.
-*   **/transactions** — Code samples showing how to use the TransactionalMap.
+    ```bash
+    docker run -p 5701:5701 hazelcast/hazelcast:4.2
+    ```
+  To see the other ways to start Hazelcast members, see 
+  [Working with Hazelcast IMDG Clusters](https://hazelcast.readthedocs.io/en/stable/getting_started.html#working-with-hazelcast-imdg-clusters)
+  section of our documentation.
+  
--- a/examples/sql/sql_async_example.py
+++ b/examples/sql/sql_async_example.py
@@ -3,6 +3,7 @@
 # for the capabilities of the SQL service. The code sample here is just to
 # demonstrate the concept of async usage of the SQL service. When in doubt,
 # use the blocking iterator method in `sql_example.py`.
+from __future__ import print_function
 
 import hazelcast
 
--- a/examples/sql/sql_async_example.py
+++ b/examples/sql/sql_async_example.py
@@ -1,0 +1,40 @@
+# One can use the SQL service in a non-blocking manner. But, such usage
+# requires more attention and is more prone to failures. See `sql_example.py`
+# for the capabilities of the SQL service. The code sample here is just to
+# demonstrate the concept of async usage of the SQL service. When in doubt,
+# use the blocking iterator method in `sql_example.py`.
+
+import hazelcast
+
+client = hazelcast.HazelcastClient()
+
+# Get and fill a map with some integers
+integers = client.get_map("integers").blocking()
+for i in range(100):
+    integers.set(i, i)
+
+# Fetch values in between (40, 50)
+result = client.sql.execute("SELECT * FROM integers WHERE this > ? AND this < ?", 40, 50)
+
+
+def on_iterator_response(iterator_future):
+    it = iterator_future.result()
+
+    def on_next_row(row_future):
+        try:
+            row = row_future.result()
+            # Process the row.
+            print(row)
+
+            # Iterate over the next row.
+            next(it).add_done_callback(on_next_row)
+        except StopIteration:
+            # Exhausted the iterator. No more rows are left.
+            pass
+
+    next(it).add_done_callback(on_next_row)
+
+
+# Request the iterator over rows and add a callback to
+# run, when the response comes
+result.iterator().add_done_callback(on_iterator_response)


hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        Returns:
+            int: Column index or :const:`COLUMN_NOT_FOUND` if a column
+            with the given name is not found.
+        """
+        check_not_none(column_name, "Column name cannot be None")
+        return self._name_to_index.get(column_name, SqlRowMetadata.COLUMN_NOT_FOUND)
+
+    def __repr__(self):
+        return "[%s]" % ", ".join(
+            map(
```
yuce, line 499: A list comprehension may be used instead of `map`.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        Returns:
+            list: The row pointed by the current position.
+        """
+        values = []
+        for i in range(self.page.column_count):
+            value = self.page.get_column_value(i, self.position)
+
+            # The column might contain user objects so we have to deserialize it.
+            # This call is no-op if the value is not Data.
+            values.append(self.deserialize_fn(value))
```
yuce, line 667: This code seems to be appropriate for a list comprehension.

yuce (changes requested):
I didn't yet finish the review, but the PR looks great overall. Some small changes may make the PR even better.

hazelcast/protocol/builtin.py
```diff
@@ -496,3 +554,206 @@ def encode(buf, value, is_final=False):
...
+    @staticmethod
+    def decode(msg, item_size, decoder):
+        frame = msg.next_frame()
+        type = FixSizedTypesCodec.decode_byte(frame.buf, 0)
+        count = FixSizedTypesCodec.decode_int(frame.buf, 1)
+
+        response = []
+        if type == ListCNFixedSizeCodec._TYPE_NULL_ONLY:
+            for _ in range(count):
+                response.append(None)
```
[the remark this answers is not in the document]
yuce: This one is WAY faster: `ls.extend([None] * count)` if it's OK to use more memory.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+          try to fetch something from the server.
+
+        Args:
+            future (Future): The response from the server for
+            the fetch request.
+        """
+        try:
+            response = future.result()
+
+            response_error = self._handle_response_error(response["error"])
```
yuce, line 1093: Just a question, is `error` key always set in `response`?

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+          (iterator, update_count etc.) fails immediately
+        - setting it to an execute response
+
+        Args:
+            future (Future):
+        """
+        try:
+            response = future.result()
+
+            response_error = self._handle_response_error(response["error"])
```
yuce, line 1146: Same question about whether `error` key always exists.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+            if response_error:
+                # There is a server-side error sent to the client.
+                self._on_execute_error(response_error)
+                return
+
+            row_metadata = response["row_metadata"]
+            if row_metadata is not None:
```
yuce, line 1153: Just a note: not setting `row_metadata` key instead of setting it to `None` may be better.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        Args:
+            error (_SqlError): The error or ``None``.
+
+        Returns:
+            HazelcastSqlError: If the error is not ``None``,
+            ``None`` otherwise.
+        """
+        if error:
+            return HazelcastSqlError(error.originating_member_uuid, error.code, error.message, None)
+        return None
```
yuce, line 1177: I think we can remove the `return None` line.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+                return
+
+            if row_metadata:
+                # Result contains the row set for the query.
+                # Set the update count to -1.
+                response = _ExecuteResponse(row_metadata, row_page, -1)
+
+                if row_page.is_last:
+                    # This is the last page, close the result.
+                    self._mark_closed()
```
yuce, line 1215: `self._mark_closed()` also acquires the lock. How about using `self._closed = True` here?

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+
+                self._execute_response.set_result(response)
+            else:
+                # Result only contains the update count.
+                response = _ExecuteResponse(None, None, update_count)
+                self._execute_response.set_result(response)
+
+                # There is nothing more we can get from the server.
+                self._mark_closed()
```
yuce, line 1224: The same comment about `_mark_closed()`

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        """Marks the result as closed."""
+        with self._lock:
+            self._closed = True
+
+    def __enter__(self):
+        # The execute request is already sent.
+        # There is nothing more to do.
+        return self
+
+    def __exit__(self, *_):
```
yuce, line 1236: Just a note: Although valid, `*_` is a bit strange.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+                None,
+            )
+
+        # Create a new, unique query id.
+        query_id = _SqlQueryId.from_uuid(connection.remote_uuid)
+
+        # Serialize the passed parameters.
+        serialized_params = []
+        for param in statement.parameters:
+            serialized_params.append(self._serialization_service.to_data(param))
```
yuce, line 1297: List comprehension opportunity.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+
+        request = sql_execute_codec.encode_request(
+            statement.sql,
+            serialized_params,
+            # to_millis expects None to produce -1
+            to_millis(None if statement.timeout == -1 else statement.timeout),
```
yuce, line 1303: How about `-1 if statement.timeout == -1 else statement.timeout*1000` instead of `to_millis( ... )`?

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+
+        Returns:
+            HazelcastSqlError: The re raised error.
+        """
+        if not connection.live:
+            return HazelcastSqlError(
+                self.get_client_id(),
+                _SqlErrorCode.CONNECTION_PROBLEM,
+                "Cluster topology changed while a query was executed: Member cannot be reached: "
+                + connection.remote_address,
```
yuce, line 1372: Using `%` syntax would be better instead of concatenating strings.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        are different schemas.
+
+        The default value is ``None`` meaning only the default search path is
+        used.
+        """
+        return self._schema
+
+    @schema.setter
+    def schema(self, schema):
+        self._schema = schema
```
yuce, line 1476: Would a regular `schema` class variable work better? Since the getter and setter don't seem to do anything special.

hazelcast/util.py
```diff
@@ -333,3 +382,79 @@ class IterationType:
 
     ENTRY = 2
     """Iterate over entries"""
+
+
+class UuidUtil(object):
```
yuce, line 387: Would it be better to name it `UUIDUtil`? I think `UUID` is used elsewhere.

yuce (changes requested):
A few more comments...

hazelcast/client.py
```diff
@@ -388,6 +389,10 @@ def __init__(self, **kwargs):
         self._invocation_service.init(
             self._internal_partition_service, self._connection_manager, self._listener_service
         )
+        self._internal_sql_service = _InternalSqlService(
+            self._connection_manager, self._serialization_service, self._invocation_service
+        )
+        self.sql = SqlService(self._internal_sql_service)
```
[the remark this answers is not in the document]
mdumandag: The reasoning is that there should be some public methods on the SQL service so that they can be used within SqlResult. I could make them protected via `_` and hide them from users but this time, when I try to use them, IDE would show warnings for access to protected members which I would like to avoid. So, SqlService wraps InternalSqlService and exposes truly public APIs. We used a similar approach in a couple of other places.

So, I could merge them but this time, we should be okay with the IDE showing warnings

hazelcast/connection.py
```diff
@@ -146,17 +146,20 @@ def add_listener(self, on_connection_opened=None, on_connection_closed=None):
...
-            member = self._load_balancer.next()
-            if member:
-                connection = self.get_connection(member.uuid)
-                if connection:
-                    return connection
+            connection = self._get_connection_from_load_balancer(should_get_data_member)
+            if connection:
+                return connection
 
         # We should not get to this point under normal circumstances.
```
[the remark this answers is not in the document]
mdumandag: Yes, let me update it, the comment only makes sense for a smart client. The copying is done to not get concurrent modification exceptions since that dict might be modified by reactor thread while iterating.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
+import logging
+import uuid
+from threading import RLock
+
+from hazelcast.errors import HazelcastError
+from hazelcast.future import Future, ImmediateFuture
+from hazelcast.invocation import Invocation
+from hazelcast.util import (
+    UuidUtil,
+    check_not_none,
```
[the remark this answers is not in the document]
mdumandag: I think it is easier to write 
```
check_not_none(sql, "SQL cannot be None")
```
than
```
if sql is None:
   raise SomeError("SQL cannot be None") 
```

If it is not a problem, I would like to leave this as it is

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+    to_millis,
+    check_true,
+    get_attr_name,
```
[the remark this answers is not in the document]
mdumandag: Well, this is a limitation due to Python 2 support. We cannot use `enum`s so we use custom classes with class variables as enum members. 

I couldn't get how a value may be mapped to different keys as long as we make sure that each enum member has a unique value. 

Note that, this is not the only place we need reverse mappings of values to keys. I wanted to write something general that would apply to all places where we use such enum-like classes.

I believe we could probably write some base metaclass that does the reverse mapping once but I think it is not a priority now. We use `get_attr_name` only on `__repr__` methods which should be fine to be less efficient.

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        self._column_types = column_types
+        self._columns = columns
+        self._is_last = last
+
+    @property
+    def row_count(self):
+        """int: Number of rows in the page."""
+        # Each column should have equal number of rows.
+        # Just check the first one.
+        return len(self._columns[0])
```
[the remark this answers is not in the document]
mdumandag: Nope, there is at least one column

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+    def get_column_value(self, column_index, row_index):
+        """
+        Args:
+            column_index (int):
+            row_index (int):
+
+        Returns:
+            The value with the given indexes.
+        """
+        return self._columns[column_index][row_index]
```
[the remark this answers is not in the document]
mdumandag: I thought about this, but in that case, we would throw something like `IndexError` which is already done by such usage. So, I didn't see an additional value

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        See Also:
+            :attr:`metadata`
+
+            :func:`SqlRowMetadata.find_column`
+
+            :attr:`SqlColumnMetadata.type`
+
+            :attr:`SqlColumnMetadata.name`
+        """
+        index = self._row_metadata.find_column(column_name)
```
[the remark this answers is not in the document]
mdumandag: I would like to stick to the API provided by Java (to throw `Column '%s' doesn't exist` kind of an error)

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        return self._row[column_index]
+
+    @property
+    def metadata(self):
+        """SqlRowMetadata: The row metadata."""
+        return self._row_metadata
+
+    def __repr__(self):
+        def mapping(column_index):
+            metadata = self._row_metadata.get_column(column_index)
```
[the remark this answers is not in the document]
mdumandag: Is this comment points to the correct place? I couldn't understand how I could throw `IndexError` or use that fact in the `__repr__` method

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        """SqlRowMetadata: Row metadata or None, if the response only 
+        contains update count."""
+
+        self.row_page = row_page
+        """_SqlPage: First page of the query response or None, if the
+        response only contains update count.
+        """
+
+        self.update_count = update_count
+        """int: Update count or -1 if row metadata or row page exist."""
```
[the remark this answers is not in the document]
mdumandag: I tried to document the instance variables, not the `__init__` method

hazelcast/sql.py
```diff
@@ -0,0 +1,1604 @@
...
+        """_SqlPage: Current page."""
+
+        self.row_count = 0
+        """int: Number of rows in the current page."""
+
+        self.position = 0
+        """int: Index of the next row in the page."""
+
+        self.is_last = False
+        """bool: Whether this is the last page or not."""
```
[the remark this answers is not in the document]
mdumandag: Same, I tried to document instance variables and this is a valid Google-style docstring. Check https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html and see the following line
```
        self.attr5 = None
        """str: Docstring *after* attribute, with type specified."""
```

hazelcast/sql.py
mdumandag: Yes, it comes from the codec. see sql_fetch_codec#decode_response

hazelcast/sql.py
mdumandag: yes

hazelcast/sql.py
mdumandag: if I understand your note correctly, you mean we shouldn't set `response["row_metadata"]` to `None`. Instead, there shouldn't be such a key in response dict. 

That requires a codec template change, and personally, I would like to leave this as it is to distinguish `None` value with no response parameter named as such cases.

If I understand your note incorrectly, please comment

hazelcast/sql.py
mdumandag: So, if I remove this, it will implicitly return `None`. I would like to be more explicit with my intent and leave this as it is

hazelcast/sql.py
mdumandag: If it is okay, I would like to leave it like this. It makes documentation more clear and it looks the same as other properties. Otherwise, this would be shown differently.

See the picture below. Not a big deal, but I would like to be consistent.

Screenshot from 2021-06-01 16-39-52

examples/sql/sql_example.py
```diff
@@ -0,0 +1,99 @@
...
+# Project a single column that fits the criterion
+result = client.sql.execute("SELECT name FROM customers WHERE age < 35 AND is_active")
+
+for row in result:
+    # Get the object with the given column name in the row
+    name = row.get_object("name")
+    print(name)
+
+
+# One can also select all fields
```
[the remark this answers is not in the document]
mdumandag: The sample is actually just below it. I just used one statement to show two different things
`SELECT * FROM customers`

hazelcast/protocol/builtin.py
```diff
@@ -496,3 +554,206 @@ def encode(buf, value, is_final=False):
...
+
+            while read_count < count:
+                bitmask = FixSizedTypesCodec.decode_byte(frame.buf, position)
+                position += 1
+
+                i = 0
+                while i < ListCNFixedSizeCodec._ITEMS_PER_BITMASK and read_count < count:
+                    mask = 1 << i
+                    if (bitmask & mask) == mask:
+                        response.append(decoder(frame.buf, position))
```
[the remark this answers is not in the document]
mdumandag: I think this one reads a bit better so I would like to leave it as it is

hazelcast/protocol/builtin.py
```diff
@@ -496,3 +554,206 @@ def encode(buf, value, is_final=False):
...
+        is_last = LE_INT8.unpack_from(msg.next_frame().buf, 0)[0] == 1
+
+        # read column types
+        column_type_ids = ListIntegerCodec.decode(msg)
+
+        # read columns
+        columns = []
+
+        for column_type_id in column_type_ids:
+            if column_type_id == SqlColumnType.VARCHAR:
```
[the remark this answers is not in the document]
mdumandag: I tried this but the end result was something uglier (to my taste :smile: )

Would it be okay to leave it like this?

hazelcast/sql.py
mdumandag: I think the current version is better because I know (from the other places), we use None for unset timeout variables. But if you think we really need to change this I can do it

examples/sql/sql_example.py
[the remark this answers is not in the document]
yuce: OK, I think removing the blank line after comment `One can also select all fields`  and the blank line between the next comment and the `with` statement below, it may make it easier to relate the comment with the `with` statement.

hazelcast/protocol/builtin.py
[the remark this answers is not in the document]
yuce: Well, apart from being idiomatic, `dict` approach is also faster. Long if/elif chains (and loops) are slow in Python, but `dict` is implemented in C. In the code below, there's quite a difference between both approaches: https://gist.github.com/yuce/cadd9b39a06bee679b75454004d520aa

Using the `dict` approach also makes it possible to get rid of the for loop and use a list comprehension which would make it even faster.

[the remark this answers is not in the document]
mdumandag: I don't think performance should be a decisive factor here. 

If you change the gist you shared with the code I pasted there, I couldn't see much difference. In the code I shared, I passed 10 as the column type (since on average we should expect to make (1 + 19)/2 comparisons) and did something pretty simple to decode a buffer. 

Note that, on the actual code, the overhead is much larger (decoding the whole row-set for a single column per column type), I believe dict or if/else approach would be identical because the actual decoding would dwarf the iteration method overhead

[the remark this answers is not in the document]
yuce: OK, fair enough. Here's how the dict approach looks like, it helps getting rid of the loop (IMO in an elegant way):
```python
# read columns
try:
    columns = [_DECODER[tid](msg) for tid in column_type_ids]
except KeyError as e:
    raise ValueError("Unknown type %s" % e.args[0])
# ...
```
There's no extra function calls for most of the codecs. `_DECODER` and the few helpers are outside the function:
```python
def _decodeNull(msg):
    frame = msg.next_frame()
    size = FixSizedTypesCodec.decode_int(frame.buf, 0)
    return [None] * size

def _decodeVarChar(msg):
  return ListMultiFrameCodec.decode_contains_nullable(msg, StringCodec.decode)

_DECODER = {
    SqlColumnType.VARCHAR: _decodeVarChar,
    SqlColumnType.BOOLEAN: ListCNBooleanCodec.decode,
    # ...
    SqlColumnType.NULL: _decodeNull,
}
```

I think at the very least we should reduce the number of `column.append`s:
```python
if column_type_id == SqlColumnType.VARCHAR:
    column = ListMultiFrameCodec.decode_contains_nullable(msg, StringCodec.decode)
elif column_type_id == SqlColumnType.BOOLEAN:
    column = ListCNBooleanCodec.decode(msg)
# ...
columns.append(column)
```

yuce (approved):

examples/sql/sql_example.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Get the index of the is_active column
+        is_active_index = row_metadata.find_column("is_active")
+
+        # Get the object with the column index
+        is_active = row.get_object_with_index(is_active_index)
+
+        print(name, age, is_active)
+
+
```
srknzl, line 71: same comment for new lines

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+        are using ``hazelcast`` or ``hazelcast-enterprise``, then you need to have
+        ``hazelcast-sql`` in the classpath. If you are using the Docker image,
+        the SQL module is included by default.
+
+    """
+
+    def __init__(self, internal_sql_service):
+        self._service = internal_sql_service
+
+    def execute(self, sql, *params):
```
srknzl, line 141: Is it ok that options is not allowed here?

hazelcast/client.py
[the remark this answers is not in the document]
srknzl: > for access to protected members which I would like to avoid.

just wondering, what is protected members? functions with starts with _?

hazelcast/protocol/builtin.py
```diff
@@ -496,3 +554,203 @@ def encode(buf, value, is_final=False):
...
+        type = FixSizedTypesCodec.decode_byte(frame.buf, 0)
+        count = FixSizedTypesCodec.decode_int(frame.buf, 1)
+
+        if type == ListCNFixedSizeCodec._TYPE_NULL_ONLY:
+            return [None] * count
+        elif type == ListCNFixedSizeCodec._TYPE_NOT_NULL_ONLY:
+            header_size = ListCNFixedSizeCodec._HEADER_SIZE
+            return [decoder(frame.buf, header_size + i * item_size) for i in range(count)]
+        else:
+            response = []
```
srknzl, line 580: you can use fixed size list for better performance. see https://stackoverflow.com/a/10617221

```python
response = [None]*count
// some lines...
response[i] = decoder(frame.buf, position)
```

hazelcast/protocol/builtin.py
```diff
@@ -496,3 +554,203 @@ def encode(buf, value, is_final=False):
...
+        msg.next_frame()
+
+        # read the "last" flag
+        is_last = LE_INT8.unpack_from(msg.next_frame().buf, 0)[0] == 1
+
+        # read column types
+        column_type_ids = ListIntegerCodec.decode(msg)
+
+        # read columns
+        columns = []
```
srknzl, line 711: same comment for using fix sized list

srknzl (commented):
Left some comments, I may continue reviewing, havent checked hazelcast/sql.py, hazelcast/util.py and tests.

hazelcast/client.py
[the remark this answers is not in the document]
mdumandag: yes

hazelcast/util.py
```diff
@@ -272,18 +276,65 @@ def next(self):
...
+        return True
+
     def _listener(self, _):
-        self._members = self._cluster_service.get_members()
+        members = self._cluster_service.get_members()
+        data_members = []
+
+        for member in members:
+            if not member.lite_member:
+                data_members.append(member)
```
srknzl, line 332: you could simplify here

```suggestion
        data_members = list(filter(lambda member: not member.lite_member, members))
```

hazelcast/util.py
```diff
@@ -312,15 +362,14 @@ def next(self):
...
-        members = self._members
+    def _next(self, members):
         if not members:
             return None
         idx = random.randrange(0, len(members))
         return members[idx]
 
 
-class IterationType:
+class IterationType(object):
```
srknzl, line 372: Just curious, what is this useful for?

hazelcast/util.py
```diff
@@ -333,3 +382,79 @@ class IterationType:
...
+
+
+else:
+    # Compatibility with Python 2
+    def int_from_bytes(buffer):
+        buffer = bytearray(buffer)
+        if buffer[0] & 0x80:
+            neg = bytearray()
+            for c in buffer:
+                neg.append(c ^ 0xFF)
```
srknzl, line 418: why not just `~c` which makes intention clearer?

hazelcast/util.py
```diff
@@ -333,3 +382,79 @@ class IterationType:
...
+        def utcoffset(self, dt):
+            return self._offset
+
+        def tzname(self, dt):
+            return None
+
+        def dst(self, dt):
+            return timedelta(0)
+
+    timezone = FixedOffsetTimezone
```
srknzl, line 446: does this minimal implementation have a drawback compared to python3?(I dont know maybe related to dst etc.)

srknzl (commented):
Left some comments. My implementation review is finished.

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+    ``partitioned`` schema. The ``partitioned`` schema is included into
+    a default search path, therefore a Map could be referenced in an
+    SQL statement with or without the schema name.
+
+    **Column resolution**
+
+    Every table backed by a Map has a set of columns that are resolved
+    automatically. Column resolution uses Map entries located on the
+    member that initiates the query. The engine extracts columns from a
+    key and a value and then merges them into a single column set.
```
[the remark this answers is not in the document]
mdumandag: Nope, the oxford comma should only be used if there are three or more terms https://en.wikipedia.org/wiki/Serial_comma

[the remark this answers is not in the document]
srknzl: I think here it is not serial comma, it is a comma needed before a conjunction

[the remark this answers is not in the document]
srknzl: https://www.grammar-monster.com/lessons/pics/comma_before_and.png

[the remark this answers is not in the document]
mdumandag: Well, the `then merges them into a single column set.` is not a full sentence, so I don't think we need `,` there. Also, Grammarly complains when I put a comma there.

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+      to lower case. A getter takes precedence over a field in case of naming
+      conflict
+    - For :class:`Portable <hazelcast.serialization.api.Portable>` objects,
+      field names used in the
+      :func:`write_portable() <hazelcast.serialization.api.Portable.write_portable>`
+      method are used to populate the column list
+
+    The whole key and value objects could be accessed through a special fields
+    ``__key`` and ``this``, respectively. If key (value) object has fields,
+    then the whole key (value) field is exposed as a normal field. Otherwise the
```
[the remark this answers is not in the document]
mdumandag: Like I said if the `__key` is a complex object, it will return it as a whole. So, you cannot access them like `age`, `name` but have to use `__key.age` and `__key.name`

hazelcast/sql.py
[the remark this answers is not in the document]
srknzl: ok I see, sorry; ignore this change then

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+
+    @property
+    def nullable(self):
+        """bool: ``True`` if the rows in this column might be ``None``,
+        ``False`` otherwise.
+        """
+        return self._nullable
+
+    def __repr__(self):
+        return "%s %s" % (self.name, get_attr_name(SqlColumnType, self.type))
```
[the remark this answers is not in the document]
mdumandag: I liked the current version more because it is simpler. And without a context, `name type True` seems really weird.

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+                self.timeout,
+                self.cursor_buffer_size,
+                self._expected_result_type,
+            )
+        )
+
+
+# These are imported at the bottom of the page to get rid of the
+# cyclic import errors.
+from hazelcast.protocol.codec import sql_execute_codec, sql_fetch_codec, sql_close_codec
```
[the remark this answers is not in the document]
mdumandag: No, they have to be imported after some classes are defined. I could put it into a middle of the file, but that doesn't look good so I think I am leaving this as it is

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+    TIMEOUT_NOT_SET = -1
+
+    TIMEOUT_DISABLED = 0
+
+    DEFAULT_TIMEOUT = TIMEOUT_NOT_SET
+
+    DEFAULT_CURSOR_BUFFER_SIZE = 4096
+
+    def __init__(self, sql):
+        self.sql = sql
```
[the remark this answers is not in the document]
mdumandag: yes

hazelcast/util.py
mdumandag: To mark this as new style class

hazelcast/util.py
mdumandag: I don't think so, this is enough to store the timezone information we have

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+
+        # Get a random Data member (non-lite member)
+        connection = self._connection_manager.get_random_connection(True)
+        if not connection:
+            # Either the client is not connected to the cluster, or
+            # there are no data members in the cluster.
+            raise HazelcastSqlError(
+                self.get_client_id(),
+                _SqlErrorCode.CONNECTION_PROBLEM,
+                "Client is not currently connected to the cluster.",
```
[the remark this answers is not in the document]
srknzl: I still believe it may be misleading, i.e when there is no connection problem -client is connected to a cluster with lite members-; it will say client is not connected to the cluster

docs/api/sql.rst
```diff
@@ -0,0 +1,4 @@
+SQL
+===========
+
+.. automodule:: hazelcast.sql
```
srknzl, line 4: does this file belong to code samples?

examples/sql/sql_example.py
```diff
@@ -0,0 +1,99 @@
...
+    row_metadata = result.get_row_metadata().result()
+
+    for row in result:
+        key = row.get_object("__key")
+        age = row.get_object("age")
+
+        print(key, age)
+
+# Parameters can also be specified for the basic query execution
+result = client.sql.execute("SELECT this FROM customers WHERE age > ?", 40)
```
srknzl, line 92: Maybe we can use an example here with more than one parameter to better show the usage.

Also, I feel like we can improve `Parameters can also be specified for the basic query execution` to maybe be `Parameters can be passed directly in the basic execution syntax`

hazelcast/sql.py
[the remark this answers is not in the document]
srknzl: My problem is with `key (value)` part. It makes the sentence complex. I would remove `(value)`

hazelcast/protocol/builtin.py
```diff
@@ -496,3 +554,203 @@ def encode(buf, value, is_final=False):
...
+            elif column_type_id == SqlColumnType.DOUBLE:
+                columns.append(ListCNDoubleCodec.decode(msg))
+            elif column_type_id == SqlColumnType.DATE:
+                columns.append(ListCNLocalDateCodec.decode(msg))
+            elif column_type_id == SqlColumnType.TIME:
+                columns.append(ListCNLocalTimeCodec.decode(msg))
+            elif column_type_id == SqlColumnType.TIMESTAMP:
+                columns.append(ListCNLocalDateTimeCodec.decode(msg))
+            elif column_type_id == SqlColumnType.TIMESTAMP_WITH_TIME_ZONE:
+                columns.append(ListCNOffsetDateTimeCodec.decode(msg))
```
srknzl, line 739: Where do you convert these to string? In the documentation, you say strings are returned.

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+    def test_execute_statement_with_timeout(self):
+        entry_count = 100
+        self._populate_map(entry_count, lambda v: Student(v, v))
+        statement = SqlStatement("SELECT age FROM %s WHERE height < 10" % self.map_name)
+        statement.timeout = 100
+        result = self.client.sql.execute_statement(statement)
+
+        six.assertCountEqual(
+            self, [i for i in range(10)], [row.get_object("age") for row in result]
+        )
```
srknzl, line 145: What is this test for? I don't why but a query with timeout does not return an error even if it is set to 0. Only setter for timeout is tested in this test

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+    def test_execute_statement_with_cursor_buffer_size(self):
+        entry_count = 50
+        self._populate_map(entry_count, lambda v: Student(v, v))
+        statement = SqlStatement("SELECT age FROM %s" % self.map_name)
+        statement.cursor_buffer_size = 3
+        result = self.client.sql.execute_statement(statement)
+
+        six.assertCountEqual(
+            self, [i for i in range(entry_count)], [row.get_object("age") for row in result]
+        )
```
srknzl, line 156: You can add a spy to fetch calls and assert its count here

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+
+
+class SqlResultTest(SqlTestBase):
+    def test_blocking_iterator(self):
+        self._populate_map()
+        result = self.client.sql.execute("SELECT __key FROM %s" % self.map_name)
+
+        six.assertCountEqual(
+            self, [i for i in range(10)], [row.get_object_with_index(0) for row in result]
+        )
```
srknzl, line 209: This is already tested several times, do we need this one?

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+        self._populate_map()
+        result = self.client.sql.execute("SELECT this FROM %s" % self.map_name)
+
+        six.assertCountEqual(
+            self, [i for i in range(10)], [row.get_object_with_index(0) for row in result]
+        )
+
+        with self.assertRaises(ValueError):
+            for _ in result:
+                pass
```
srknzl, line 221: Is this behaviour specified anywhere in documentation/api documentation? there is one in iterator() method but it is for future producing iterator, not blocking iterator

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+
+    def test_blocking_iterator_with_multi_paged_result(self):
+        self._populate_map()
+        statement = SqlStatement("SELECT __key FROM %s" % self.map_name)
+        statement.cursor_buffer_size = 1  # Each page will contain just 1 result
+        result = self.client.sql.execute_statement(statement)
+
+        six.assertCountEqual(
+            self, [i for i in range(10)], [row.get_object_with_index(0) for row in result]
+        )
```
srknzl, line 231: why we need this and how it is different than test_execute_statement_with_cursor_buffer_size

examples/sql/sql_example.py
```diff
@@ -0,0 +1,99 @@
...
+
+for row in result:
+    customer = row.get_object("this")
+    print(customer)
+
+# Query can be closed explicitly
+result.close().result()
```
srknzl, line 99: Why dont we have a code sample for iterator() based approach? (future based)

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+
+        rows = []
+        for row_future in iterator:
+            try:
+                row = row_future.result()
+                rows.append(row.get_object_with_index(0))
+            except StopIteration:
+                break
+
+        six.assertCountEqual(self, [i for i in range(10)], rows)
```
srknzl, line 300: I don't think we need `cursor_buffer_size = 1` versions of several tests. Just one of cursor_buffer_size test to test pagination is enough.

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+    def test_get_row_metadata(self):
+        self._populate_map(value_factory=str)
+        result = self.client.sql.execute("SELECT * FROM %s" % self.map_name)
+        row_metadata = result.get_row_metadata().result()
+        self.assertEqual(2, row_metadata.column_count)
+        columns = row_metadata.columns
+        self.assertEqual(SqlColumnType.INTEGER, columns[0].type)
+        self.assertEqual(SqlColumnType.VARCHAR, columns[1].type)
+        self.assertTrue(columns[0].nullable)
+        self.assertTrue(columns[1].nullable)
```
srknzl, line 345:
```suggestion
    def test_get_row_metadata(self):
        self._populate_map(value_factory=str)
        result = self.client.sql.execute("SELECT * FROM %s" % self.map_name)
        row_metadata = result.get_row_metadata().result()
        self.assertEqual(2, row_metadata.column_count)

        key_column_index = row_metadata.find_column("__key")
        this_column_index = row_metadata.find_column("this")

        key_column = row_metadata.get_column(key_column_index)
        this_column = row_metadata.get_column(this_column_index)

        self.assertEqual(SqlColumnType.INTEGER, key_column.type)
        self.assertEqual(SqlColumnType.VARCHAR, this_column.type)
        self.assertTrue(key_column.nullable)
        self.assertTrue(this_column.nullable)
```

to test public API and to be sure of column indexes

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+        self._populate_map()
+        statement = SqlStatement("SELECT this FROM %s" % self.map_name)
+        statement.cursor_buffer_size = 1  # so that it doesn't close immediately
+
+        with self.assertRaises(RuntimeError):
+            with self.client.sql.execute_statement(statement) as result:
+                for _ in result:
+                    raise RuntimeError("expected")
+
+        self.assertIsInstance(result.close(), ImmediateFuture)
```
srknzl, line 387: Is it hard in python to add a spy to close() and assert that it is called? I know it is easy in Node.js

tests/unit/sql_test.py
```diff
@@ -0,0 +1,302 @@
...
+    def get_message_type(self, invocation):
+        return LE_INT.unpack_from(invocation.request.buf, _OUTBOUND_MESSAGE_MESSAGE_TYPE_OFFSET)[0]
+
+    def set_future_result_or_exception(self, value, message_type):
+        for invocation in self.invocation_registry.values():
+            if self.get_message_type(invocation) == message_type:
+                if isinstance(value, Exception):
+                    invocation.future.set_exception(value)
+                else:
+                    invocation.future.set_result(value)
```
srknzl, line 244: I don't like this approach. As a result of `close`, `fetch` calls you get a future, why not directly set_result of it but use invitation registry. I think it is unnecessary complexity.

tests/unit/sql_test.py
```diff
@@ -0,0 +1,302 @@
...
+
+        # initiate the fetch request
+        future = next(i)
+
+        self.set_fetch_response_with_error()
+
+        with self.assertRaises(HazelcastSqlError) as cm:
+            future.result()
+
+        self.assertEqual(_SqlErrorCode.PARSING, cm.exception._code)
```
srknzl, line 163: Same here I feel like this tests is not very useful due to the same reason

srknzl (commented):
Finished my review (impl, tests, doc, code samples included), left some comments for you

hazelcast/sql.py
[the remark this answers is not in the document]
mdumandag: will change it to `key (or value)`

tests/integration/backward_compatible/sql_test.py
mdumandag: 0 is not an erroneous case. 0 means there is no timeout. This test is a valid one because the IMDG SQL engine will timeout your queries (based on the value of this) if your query is a long-running one. There is no deterministic way to get timeout exceptions in queries but if you fill your map with enough items and run somewhat complex query, you might get timeouts.

As for the test, I don't think it is an unnecessary one

tests/integration/backward_compatible/sql_test.py
mdumandag: This is an integration test where we use the real code, not the mocked one. I am not planning to add mocking to this test but as long as the server is not faulty (respecting the cursor buffer size), fetch will be called.

tests/integration/backward_compatible/sql_test.py
mdumandag: I have to use mock objects and we don't do it in integration tests. possible, but not that good looking. we already test lots of cases in unit tests so I am planning to not do it here

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+        self._validate_result(result, expected_type, value_factory)
+
+    def _validate_result(self, result, expected_type, factory):
+        for row in result:
+            key = row.get_object("__key")
+            expected_value = factory(key)
+            row_metadata = row.metadata
+
+            self.assertEqual(2, row_metadata.column_count)
+            column_metadata = row_metadata.get_column(1)
```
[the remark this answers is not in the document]
mdumandag: Some of these weird-looking accesses are for code coverage. Code coverage is really important in a highly dynamic language like Python, so I want to leave these as it is for this reason.

tests/integration/backward_compatible/sql_test.py
```diff
@@ -0,0 +1,590 @@
...
+        self._populate_map(value_factory=value_factory)
+        result = self.client.sql.execute("SELECT __key, this FROM %s" % self.map_name)
+        self._validate_result(result, SqlColumnType.OBJECT, value_factory)
+
+    def test_null_only_column(self):
+        self._populate_map()
+        result = self.client.sql.execute(
+            "SELECT __key, CAST(NULL AS INTEGER) FROM %s" % self.map_name
+        )
+        self._validate_result(result, SqlColumnType.INTEGER, lambda _: None)
```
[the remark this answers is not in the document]
mdumandag: For code coverage

tests/unit/sql_test.py
```diff
@@ -0,0 +1,302 @@
...
+        self.assertEqual(_SqlErrorCode.CANCELLED_BY_USER, cm.exception._code)
+
+    def test_close_when_close_request_fails(self):
+        future = self.result.close()
+        self.set_close_error(HazelcastSqlError(None, _SqlErrorCode.MAP_DESTROYED, "expected", None))
+
+        with self.assertRaises(HazelcastSqlError) as cm:
+            future.result()
+
+        self.assertEqual(_SqlErrorCode.MAP_DESTROYED, cm.exception._code)
```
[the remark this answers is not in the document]
mdumandag: Required for code coverage

tests/unit/sql_test.py
```diff
@@ -0,0 +1,302 @@
...
+
+        # initiate the fetch request
+        future = next(i)
+
+        self.set_fetch_error(RuntimeError("expected"))
+
+        with self.assertRaises(HazelcastSqlError) as cm:
+            future.result()
+
+        self.assertEqual(_SqlErrorCode.GENERIC, cm.exception._code)
```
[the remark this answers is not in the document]
mdumandag: Same, code coverage

tests/unit/sql_test.py
mdumandag: We are mocking things like invocation service, connection manager. etc. to use the code at the internal SQL service. This is a simulation of the server response, if we were to set the result of the future directly, I think that would be cheating :smile:

tests/unit/sql_test.py
mdumandag: Same, code coverage for error cases

tests/integration/backward_compatible/sql_test.py
srknzl: I think the timeout feature is not working yet. Try to set it to 1 and populate map with 1000000 items, then make a complex query; I don't think it will give an error.

tests/integration/backward_compatible/sql_test.py
srknzl: You don't need to mock to spy in Node.js. Does python have such feature? See spies in sinon package.

tests/integration/backward_compatible/sql_test.py
srknzl: Similarly spy should require mocking, you can spy a method without changing its behaviour in Node.js

tests/integration/backward_compatible/sql_test.py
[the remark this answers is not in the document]
srknzl: however, you cannot be sure index 1 will stay index of `this` in future versions of IMDG

How does it increase code coverage? I use more methods in my suggestion

tests/integration/backward_compatible/sql_test.py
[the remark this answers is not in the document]
srknzl: Which line(s) does this test cover, but not any other test? I think `CAST(NULL AS INTEGER)` is weird and probably won't be used at all. I don't think we should add tests just for coverage

tests/unit/sql_test.py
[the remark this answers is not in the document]
srknzl: I would rather test it like this:

Mock SQL service's close method to raise an error. Call close() on result, assert the same error is thrown. This way we also test that close method of a SQL result calls/related to SQL service's close method

tests/unit/sql_test.py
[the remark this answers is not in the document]
srknzl: similar here. 
I would rather test it like this(to make it less trivial and more useful):

Mock SQL service's fetch method to raise an error. Make result start fetch request, assert the same error is thrown. This way we also test that _fetch_next_page method of a SQL result calls/related to SQL service's fetch method

tests/unit/sql_test.py
srknzl: No it would be simpler mocking in my opinion. Mocking is fine in unit testing, adding invocation registry to the table adds complexity to the mocking we try to do.

srknzl: But if you feel this is simpler for you it's ok

tests/integration/backward_compatible/sql_test.py
mdumandag: From the client's PoV, a timeout error is no different than an error response from the server. I don't think we need a separate test for it. As for this test, I don't see a problem keeping it even if the server is not working (yet). We could add an extra test for time outing a query later. Think of this test as setting timeout to something on client and seeing how to server reacts

srknzl: another point is that this test is testing server behaviour(as you said `Think of this test as setting timeout to something on client and seeing how to server reacts`), therefore it is unnecessary for this reason too. As I said above, the only interesting thing left in this test from client's PoV is timeout's setter which is obviously not the aim of this test.

tests/integration/backward_compatible/sql_test.py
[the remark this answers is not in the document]
mdumandag: See `ListCNFixedSizeCodec`

```py
        if type == ListCNFixedSizeCodec._TYPE_NULL_ONLY:
            return [None] * count
```

There is no way I could find to test the `else:` branch with the server though, in the same codec for IMDG 4.2.

tests/unit/sql_test.py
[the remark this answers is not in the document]
mdumandag: Mocking SQL service is not a good idea since what we are trying to test is its code

tests/integration/backward_compatible/sql_test.py
srknzl: Actually it even cannot test server's behaviour due to the fact that it is hard to get a timeout error(I tried but could not get one)

mdumandag: I could even leave this here just to test Python's setter, even if this is totally unused in the server

tests/unit/sql_test.py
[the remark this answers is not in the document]
srknzl: You are testing close() behaviour of SQL result here. You will just mock close method of SQL service just for this test. You are actually doing it but via  set_close_error, with invocation registry(I am fine if it is ok for you as I said in another comment)

tests/integration/backward_compatible/sql_test.py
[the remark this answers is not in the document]
srknzl: How can someone add a null value to a map? I think it is impossible with map methods. I think due to this reason you used `CAST(NULL AS INTEGER)`, which I find weird. If a way is introduced in the future to add null values to a map via SQL, I would rather change this test later.

tests/integration/backward_compatible/sql_test.py
srknzl: Then, I would add a seperate test(and delete this test) for testing the setter with different inputs, not just with one input like here. Because you cannot test this logic with one input:

```python
@timeout.setter
def timeout(self, timeout):
    if timeout < 0 and timeout != SqlStatement.TIMEOUT_NOT_SET:
        raise ValueError("Timeout must be non-negative or -1, not %s" % timeout)

    self._timeout = timeout
```

tests/integration/backward_compatible/sql_test.py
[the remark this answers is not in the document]
mdumandag: This is how the fixed-size columns are encoded in Java, so I assume it should be possible now (or in the future), and clients should work in the future. So, I find a way to test some part of it with IMDG 4.2, and I think this is valuable. I also spent a considerable amount of time to get a response for the `else` part with 4.2, but couldn't succeeded. So, I would test it now( as long as I can)

tests/integration/backward_compatible/sql_test.py
mdumandag: I said even if totally unused :smile: and it is not unused see `SqlInternalService#execute` in Java

tests/integration/backward_compatible/sql_test.py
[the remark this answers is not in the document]
srknzl: I think your approach is a workaround, since no null data is not stored in a map, but you add an extra expression to your query. With SQL we are querying map data, that's why I mentioned adding null values to a map.

tests/integration/backward_compatible/sql_test.py
srknzl: You intend to test server then. Why you test server side code?

srknzl: As I said we even cannot test server side code, since it does not throw timeout error anyhow

tests/integration/backward_compatible/sql_test.py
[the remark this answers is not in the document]
mdumandag: Yes, I know it is a workaround. Like I said, Java encodes columns like this and on the client-side, I try to decode it. Maybe in the future, we will add this feature and I would like the client to work correctly, even now. 

So, I am not thinking about removing this. As I said, if you can find a workaround to test the `else` part too, I could happily add a test for it

tests/integration/backward_compatible/sql_test.py
mdumandag: I am not sure why we are debating this much in this simple test but I see it as valuable, so if it is not a BIG problem for you, let's leave it like this

tests/integration/backward_compatible/sql_test.py
mdumandag: I added this to be symmetrical with the future producing iterator tests, yes it is almost the same (apart from the values etc.) but I would like to keep it so that we could know that we are testing it for both iteration methods

tests/integration/backward_compatible/sql_test.py
mdumandag: Different iterators might behave differently. So, it makes sense to test them with the same set of inputs

tests/integration/backward_compatible/sql_test.py
mdumandag: To make the tests symmetrical. I would like to have a standard way of testing things even though this requires a little bit of duplication.

hazelcast/sql.py
```diff
@@ -0,0 +1,1593 @@
...
+
+    def __init__(self, row_metadata, row):
+        self._row_metadata = row_metadata
+        self._row = row
+
+    def get_object(self, column_name):
+        """Gets the value of the column by column name.
+
+        Column name should be one of those defined in :class:`SqlRowMetadata`,
+        case-sensitive. You may also use :func:`SqlRowMetadata.find_column` to
```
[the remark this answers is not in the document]
mdumandag: I think the current one is more clear, we don't need to make it more complex

tests/unit/sql_test.py
```diff
@@ -0,0 +1,302 @@
+import itertools
```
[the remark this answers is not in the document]
mdumandag: I will add some more unit tests, but like I said, most of the functionality is tested in the integration tests

tests/integration/backward_compatible/sql_test.py
mdumandag: changed the query to be sure of column indexes instead of changing the test as you suggested, again due to code coverage. I test something I don't use in other tests here

mdumandag: Did the same thing to other places to be sure of column indexes

examples/sql/sql_example.py
```diff
@@ -0,0 +1,99 @@
+from __future__ import print_function
```
srknzl, line 1: why don't we have this in async example?

hazelcast/sql.py
```diff
@@ -1481,14 +1495,15 @@ def parameters(self):
 
     @parameters.setter
     def parameters(self, parameters):
+        check_true(isinstance(parameters, list), "Parameters must be a list")
         if not parameters:
             self._parameters = []
         else:
             self._parameters = list(parameters)
```
srknzl: since we are sure that `parameters` is a list we can simplify here 
```suggestion
        self._parameters = parameters
```

hazelcast/sql.py
```diff
@@ -1542,7 +1559,8 @@ def expected_result_type(self):
 
     @expected_result_type.setter
     def expected_result_type(self, expected_result_type):
-        check_not_none(expected_result_type, "Expected result type cannot be None")
+        # Ignore the result, we call this method just to type check.
+        try_to_get_enum_value(expected_result_type, SqlExpectedResultType)
         self._expected_result_type = expected_result_type
```
srknzl: cant we do

```suggestion
        self._expected_result_type = try_to_get_enum_value(expected_result_type, SqlExpectedResultType)
```

tests/integration/backward_compatible/sql_test.py
```diff
@@ -151,9 +164,14 @@ def test_execute_statement_with_cursor_buffer_size(self):
         statement.cursor_buffer_size = 3
         result = self.client.sql.execute_statement(statement)
 
-        six.assertCountEqual(
-            self, [i for i in range(entry_count)], [row.get_object("age") for row in result]
-        )
+        with patch.object(result, "_fetch_next_page", wraps=result._fetch_next_page) as patched:
```
srknzl, line 167: not sure but looks like wraps is not a keyword argument on patch.object https://docs.python.org/3/library/unittest.mock.html#patch-object

```suggestion
        with patch.object(result, "_fetch_next_page") as patched:
```

tests/unit/sql_test.py
```diff
@@ -282,11 +281,11 @@ def test_statement_cursor_buffer_size(self):
...
-        invalid_inputs = [0, -10, -99999]
+        invalid_inputs = [0, -10, -99999, "hey", None, 1.0]
 
         for invalid in invalid_inputs:
             statement = SqlStatement("something")
-            with self.assertRaises(ValueError):
+            with self.assertRaises((ValueError, AssertionError)):
                 statement.cursor_buffer_size = invalid
 
     def test_statement_expected_result_type(self):
```
srknzl, line 291: I think 'ROWS' is also valid. Maybe we can add that too. BTW I think currently passing 'ROWS' to the expected result setter does not throw an error in setter but causes a problem when executing SQL statement. This because, as I see, try to get enum method accepts string versions of enums but it converts them to numbers.

tests/integration/backward_compatible/sql_test.py
mdumandag: It is one of the kwargs

srknzl (approved):

PR #409, opened 2021-06-03
 [API-305] Add SQL documentation 

mdumandag:
Added similar documentation to the IMDG Reference Manual
for the SQL feature. Modified examples and some parts of
the documentation to make it suit the Python client.

### DIFF
--- a/docs/using_python_client_with_hazelcast_imdg.rst
+++ b/docs/using_python_client_with_hazelcast_imdg.rst
@@ -1530,9 +1530,9 @@ Note that, the following types are returned as strings, with the following
 formats.
 
 - ``DATE`` with the ``YYYY-MM-DD`` format.
-- ``TIME`` with the ``HH:MM:SS[.mmmmmm]`` format.
+- ``TIME`` with the ``HH:MM:SS[.ffffff]`` format.
 - ``TIMESTAMP`` with the ``YYYY-MM-DDTHH:MM:SS[.ffffff]`` format.
-- ``TIMESTAMP_WITH_TIME_ZONE`` with the ``YYYY-MM-DDTHH:MM:SS.ffffff+HH:MM[:SS[.ffffff]]``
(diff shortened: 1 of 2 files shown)


docs/using_python_client_with_hazelcast_imdg.rst
```diff
@@ -1321,6 +1321,304 @@ the entry processor in the ``Map`` methods. See the following example.
...
+TIMESTAMP_WITH_TIME_ZONE str
+OBJECT                   Any Python type
+======================== ===============
+
+Note that, the following types are returned as strings, with the following
+formats.
+
+- ``DATE`` with the ``YYYY-MM-DD`` format.
+- ``TIME`` with the ``HH:MM:SS[.mmmmmm]`` format.
+- ``TIMESTAMP`` with the ``YYYY-MM-DDTHH:MM:SS[.ffffff]`` format.
```
srknzl, line 1509: why ffffff instead of you used mmmmmm above? same for timestamp with timezone

docs/using_python_client_with_hazelcast_imdg.rst
```diff
@@ -1321,6 +1321,304 @@ the entry processor in the ``Map`` methods. See the following example.
...
+- ``TIMESTAMP_WITH_TIME_ZONE`` with the ``YYYY-MM-DDTHH:MM:SS.ffffff+HH:MM[:SS[.ffffff]]``
```
srknzl, line 1510: I don't think ffff after in [:SS[.ffffff]] is possible here since offset is in seconds

docs/using_python_client_with_hazelcast_imdg.rst
mdumandag: `ffffff` is used in the datetime module documentation. so I changed `mmmmmm` into it.

srknzl: I think it is good to be consistent regarding second precision letter

mdumandag: yes, we use `ffffff` everywhere now

srknzl (approved):

PR #410, opened 2021-06-07
[API-464] Show the context manager support for transactions in code samples and…

### LINKED ISSUE #401: Use contextmanager for the transaction code samples and documentation
We have support for the contextmanager for the transactions but we are not using it for our code samples or documentation.

We should be using it to make it more readable and familiar with Python developers.

mdumandag:
… documentation

We do have a context manager support for transactions but we were
not showing it in our code samples or documentation. As a result
of that, users were unaware of this feature. We even had a call
with one user who implemented a simple context manager for
Hazelcast transactions himself.

So, with this PR, we are now showing that we have support for it
in our code samples and documentation.

closes #401

### DIFF
--- a/examples/transactions/transaction_basic_example.py
+++ b/examples/transactions/transaction_basic_example.py
@@ -2,17 +2,11 @@ import hazelcast
 import time
 
 client = hazelcast.HazelcastClient()
-transaction = client.new_transaction(timeout=10)
-try:
-    transaction.begin()
-    transactional_map = transaction.get_map("transaction-map")
-    print("Map: {}".format(transactional_map))
+
+with client.new_transaction(timeout=10) as transaction:
+    transactional_map = transaction.get_map("transactional-map")
+    print("Created map:", transactional_map)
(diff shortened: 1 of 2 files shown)


srknzl (approved):

examples/transactions/transaction_basic_example.py
```diff
@@ -2,17 +2,11 @@
...
+    print("Created map:", transactional_map)
 
     transactional_map.put("1", "1")
     time.sleep(0.1)
     transactional_map.put("2", "2")
-
-    transaction.commit()
-except Exception as ex:
-    transaction.rollback()
-    print("Transaction failed! {}".format(ex.args))
```
srknzl, line 18: Just curious, can users add exception handler like this with context manager approach?

mdumandag: they can do something like 
```py
with ... as smth:
    try:
        ...
    except ..:
       ...
```

PR #412, opened 2021-06-11
[API-466] Fire membership events when HotRestart is enabled

### LINKED ISSUE #371: [TRACKING ISSUE] Fire membership events when persistence enabled

See https://github.com/hazelcast/hazelcast/pull/18245 for details.

---

When persistence enabled, client does not fire membership
events because members starts with same uuid's.

With this pr, we make sure that when cluster restarted,
we always update the local memberlist with empty member list
first.
This way, we will be able to fire member removed/added events
even if member uuid's does not change.

Note that if cluster uuid does not change, we will not fire
any event. This happens client disconnected and connected
back to the same cluster.
Note that cluster uuid is not preserved on hotrestart with
presistence.

There was already a test to verify this behaviour but it
was working wrong.
"ClientHotRestartTest extends ClientClusterRestartEventTest"
In this pr, we make sure that the test uses same address
for the restarted member to test this correctly.

fixes https://github.com/hazelcast/hazelcast/issues/18234

mdumandag:
With this change, we now clear the memberlist when the cluster changes,
instead of just clearing the memberlist version so that, we can
fire events when a member with HotRestart enabled restarts with
the same UUID. Added a test that verifies this behavior.

Also, removed an unnecessary usage of the `clear_member_list_version`
method.

closes #371

### DIFF
--- a/hazelcast/listener.py
+++ b/hazelcast/listener.py
@@ -218,7 +218,6 @@ class ClusterViewListenerService(object):
         if self._listener_added_connection:
             return
 
-        self._cluster_service.clear_member_list_version()
         self._listener_added_connection = connection
         request = client_add_cluster_view_listener_codec.encode_request()
         invocation = Invocation(
--- a/tests/integration/backward_compatible/cluster_test.py
+++ b/tests/integration/backward_compatible/cluster_test.py
@@ -4,7 +4,7 @@ import tempfile
 from hazelcast import HazelcastClient, six
 from hazelcast.util import RandomLB, RoundRobinLB
 from tests.base import HazelcastTestCase
-from tests.util import set_attr, random_string, event_collector
+from tests.util import set_attr, random_string, event_collector, mark_client_version_at_least
 
 
 class ClusterTest(HazelcastTestCase):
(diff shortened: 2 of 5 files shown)


hazelcast/listener.py
```diff
@@ -218,7 +218,6 @@ def _try_register(self, connection):
         if self._listener_added_connection:
             return
 
-        self._cluster_service.clear_member_list_version()
```
mdumandag, line 221: This was being called again, in a few lines below

tests/integration/backward_compatible/cluster_test.py
```diff
@@ -213,6 +213,35 @@ def assertion():
         self.assertEqual(1, len(members))
         self.assertEqual(member.uuid, str(members[0].uuid))
 
+    def test_when_member_started_with_the_same_address(self):
+        mark_client_version_at_least(self, "4.2")
```
srknzl, line 217: I think I forgot to add this check in Node.js client 

https://github.com/hazelcast/hazelcast-nodejs-client/pull/896/files#diff-0bf3994a78093774f530d1d4e4521bc0da29c41e1a476312f62735f136065523R84

version check is necessary since we fixed this in 4.2 right?

I guess my test in Nodejs will fail in the next backward compat tests when a new imdg is about to be released. We need to add the version check to Nodejs too, right?

srknzl (approved):
lgtm thanks

tests/integration/backward_compatible/cluster_test.py
mdumandag: Yes, we probably need to add this to the Node.js client as well

PR #413, opened 2021-06-15
Fix invocation retry logic

### LINKED ISSUE #375: [TRACKING ISSUE] Fix ClientInvocation Retry

See https://github.com/hazelcast/hazelcast/pull/18309 for details.

---

Bug Description:
There seems to be multiple problems which is caused by single problem
we allow multiple threads to retry/change state of single invocation.
Example:
A connection is closed. CleanResourcesTask checks periodically and
handles invocation with connection closed.
   https://github.com/hazelcast/hazelcast/blob/04e9df4c2f406238267a8d45fbc9ae476b40c7d5/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientInvocationServiceImpl.java#L290-L301

This continues and schedules a retry here
    https://github.com/hazelcast/hazelcast/blob/04e9df4c2f406238267a8d45fbc9ae476b40c7d5/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientInvocation.java#L305-L315

Before this task gets a chance to run, CleanResoursTask can wakeup
check the same connection and schedule a new retry.

As a solution I first, try to sync them on `deRegisterInvocation(callid)`
,but this does not work. The idea was to only one that can
deRegisterInvocation will retry. Also a response/exception from
remote will notify if they can deregister the invocation.
One problem is correlation id is changed by the retried task, and
other threads are trying to deregister by reading that correlationId.
In response path, we have correlation id coming from remote but in
connection closed case, we don't have one to use(instead of reading
from invocation itself)

Another problem is how we handle connection close. Lets say a member
is about the close and we send an invocation to it. It can send,
ClusterPassive, InstanceNotActive like exceptions. And it closes the
connection after that. We have two cases to handle on the client side.
Lets say we have detected that connection is closed again in
CleanResoursesTask, and we are about to notify the invocation.
In the mean time we got InstanceNotActiveException as well, and we
retried it on another connection/with new call id.
Notify for connection close deregisters the new retry which should
not happen.

I have noticed we had similar problems on the core side.
#7270

I have investigate the solutions and core codebase
#9296
#9303

They are not suitable for the client. Especially connection close
case is different. On the member, we have MemberLeftException
which is guarded by `MemberListVersion` which are not applicable
to the client. So for the client, solution grew around `connection`
naturally.

Solution:
We coordinate all threads that wants to notify the invocation.
We achieve synchronization of different threads via `sentConnection`
field sentConnection starts as null.
It is set to a non-null when `connection` to be sent is determined.
It is set to null when a response/exception is going to be notified.
It is trying to be compared and set to null on connection close case
with dead connection, to prevent invocation to be notified which is
already retried on another connection.
Only one thread that can set this to null can be actively notify
a response/exception.

This way we make sure that only one thread changes the states
(changing correlation id/deregistration of invication).

fixes #18062

mdumandag:
In the current code-base, we could retry invocations in the following
cases.

- In the `_do_invoke` function, when a target connection for the
invocation cannot be determined. This may happen in any other thread
than the reactor thread. But, in this case, a message is not sent
from the client (even, a message is not even put into one of the write
queue of a connection). Hence, the `sent_connection` of such invocations
is `None`.

- When an exceptional response is received from the server, which
always happen in the reactor thread.

- When a connection is closed, and we detect it in the clean resources
timer, which also always happens in the reactor thread.

Under this scenario, it was not possible for an invocation to
be retried multiple times, once for the exceptional response and one
for the connection close case concurrently, like the Java client.

But, there were a couple of problematic scenarios.

- When we schedule an invocation for retry, we were not resetting
the `sent_connection`. So, if the `sent_connection` is closed in
the meanwhile, it was possible to schedule
`invocation_retry_pause / clean_resources_timer_period` retries.
Note that, an invocation retry which will happen in the future
does not care about its last `sent_connection`.

- If we schedule an invocation for a retry and somehow, its result
is set in the meanwhile, while we are waiting for the
`invocation_retry_pause`, we will retry the invocation anyway. Now,
we check whether or not an invocation is done, before retrying it.

To tackle 1, we reset the `sent_connection` before scheduling the retry.

To tackle 2, we check the invocation's future before retrying.

closes #375

### DIFF
--- a/hazelcast/invocation.py
+++ b/hazelcast/invocation.py
@@ -270,8 +270,13 @@ class InvocationService(object):
             self._complete_with_error(invocation, error)
             return
 
-        invoke_func = functools.partial(self._do_invoke, invocation)
+        invocation.sent_connection = None
+        invoke_func = functools.partial(self._retry_if_not_done, invocation)
         self._reactor.add_timer(self._invocation_retry_pause, invoke_func)
+
+    def _retry_if_not_done(self, invocation):
+        if not invocation.future.done():
+            self._do_invoke(invocation)
 
     def _should_retry(self, invocation, error):
         if invocation.connection and isinstance(error, (IOError, TargetDisconnectedError)):


srknzl (approved):

mdumandag:
merging this to start the SOAK tests in a timely manner. If there are any comments about it, we can discuss it (and restart the SOAK tests, if necessary)

PR #415, opened 2021-06-15
Update HZ version used in README and SOAK tests

mdumandag:
Set the version to match the client version (4.2).

Also, updated the SOAK test config so that one can easily
switch between multicast and TCP/IP configs when on does not
work.


README.rst
```diff
@@ -49,7 +49,7 @@ images <https://hub.docker.com/r/hazelcast/hazelcast/>`__.
 
 .. code:: bash
 
-   docker run -p 5701:5701 hazelcast/hazelcast:4.1
+   docker run -p 5701:5701 hazelcast/hazelcast:4.2
```
srknzl, line 52: We may update this one too: 

just above https://github.com/hazelcast/hazelcast-python-client/blob/master/docs/using_python_client_with_hazelcast_imdg.rst#logging-configuration
in Note:

 Note:: Statistics sent by Hazelcast Python client 4.0 are compatible
    with Management Center 4.0. Management Center 4.2020.08 and newer
    versions will be supported in version 4.1 of the client.

we may say `will be supported in 4.1 or newer versions of the client`

srknzl (approved):

README.rst
mdumandag: changed it to `later versions of the client` since we have no plan to do it in the near future

PR #418, opened 2021-06-21
Backward compatibility related test fixes

mdumandag:
There were two problems identified in the backward compatibility
tests.

- The `reliable_topic_test` was added to the client in v4.1 and marked
correctly. However, there were no guards on the imports. So, it failed
against client v4.0. We added import guards for problematic ones.

- In `test_async_reconnect_mode`, we added the listener after restarting
the member. In some situations, we might connect member back, even
before adding the listener, and miss the event. Now, the listener is
added before restarting the member.


srknzl (approved):
lgtm

PR #419, opened 2021-06-21
Add reliable topic configs only for 4.1+ clients

mdumandag:
In the `ReliableTopicTest`, we should only add the reliable
topic config elements for 4.1+ clients.

Note that, this change is required because the unittest will skip
tests after running the setUpClass method.


srknzl:
Is this really necessary? Did you get an error? I did not see a run with error after the last one. 

As far as I see from python docs "Skipped classes will not have setUpClass() or tearDownClass() run." right below the section https://docs.python.org/3/library/unittest.html#unittest.SkipTest

mdumandag:
This is actually required since we use nosetests as the test runner. See https://github.com/nose-devs/nose/issues/946

srknzl (approved):

PR #436, opened 2021-07-26
Fix frame reader

### LINKED ISSUE #434: Read in map extremely slow when dealing with large Payload
Hi, I kind of have a performance issue when using the Python client to retrieve very large objects (16Mbytes protobuf object for example).

There is no problem when setting the value in the cluster it takes a reasonable amount of time, but when it comes to reading it back it takes between 7 and 10 seconds.

I investigated this performance issue, my first idea was that the cluster member had difficulties to give back such a large item. That is why I implemented the same code using the Java (Scala) client: the performance issue does not appear and the reading less time than writing and is really short. In Scala I have the desired behaviour.

I then thought of a performance issue with Python, I implemented in Python using the requests module a way of writing and getting objects in a Map using the REST API. Again, with this solution, it takes more time than using the python-client for writing but there is no performance issue when reading an object.

This table compares the performance of these implementations (each figure is a mean of 100 measurements) :
| Protobuf object size | Python write | Python read | Scala write | Scala read | Python REST API write | Python REST API read |
|----------------------|--------------|-------------|-------------|------------|-----------------------|----------------------|
| 59 bytes             | 1ms          | 0.75ms      | 0.74ms      | 0.44ms     |                       |                      |
| 4 Megabytes          | 86ms         | 8681ms      | 68ms        | 20ms       |                       |                      |
| 16 Megabytes         | 337ms        | 8834ms      | 290ms       | 82ms       | 389ms                 | 155ms                |

Is there any reason why the reading in a Python Map takes more than 8 sec (!!!) to get a 16Mo object ? 

I tried to dig into the source code of the python client without finding somehing relevant. The delay seems to happen within the _send method of InvocationService.

Here is a sample code to reproduce the issue
```python
import time
import hazelcast
import numpy as np
import pandas as pd


MAP_NAME = "source_debug"

def random_picture():
    image = np.random.random((1000, 1000))
    picture = image.tobytes()
    return picture

def put_in_map(map, id, content):
    b = content
    start = time.time()
    map.set(id, bytearray(b))
    stop = time.time()
    print("Storing key({}) done in {}ms".format(len(b),
                                                   (stop - start) * 1000))
    return len(b), (stop -start) * 1000


def get_and_assert(map, id, content):
    start = time.time()
    b = bytes(map.get(id))
    stop = time.time()
    assert b == content
    print("Getting key ({}) done in {}ms".format(len(b),
                                                    (stop - start) * 1000))
    return (stop - start) * 1000


hazelcast = hazelcast.HazelcastClient()
map = hazelcast.get_map(MAP_NAME).blocking()

speed_picture = random_picture()

put_in_map(map, "test1", speed_picture)
get_and_assert(map, "test1", speed_picture)
put_in_map(map, "camera", speed_picture)
get_and_assert(map, "camera", speed_picture)

```

**TL;DR;** Python client takes more than 8 seconds to retrieve a value from a Hazelcast Map, whereas using the REST API in Python, or using the Java client, the same operation takes around 100ms

mdumandag:
In the frame reader, we were storing the frame size including the
header (frame size + flags which is 6 bytes). Also, after reading
the frame size from the header, we were incrementing the bytes_read.

That approach works for small frames, that we can read with one go,
but was not working for frames that are large. For the same frame,
the next time we call read_frame, the length of the reader (the data
we haven't read from the buffer) was equal to frame size, but the
variable frame_size was equal to `frame_size + HEADER_SIZE`. So,
we were not reading the data that was actually there due to some
faulty length check.

We were able to read the data that is in the buffer after some long
time, when the response for the client ping request comes as it
was incrementing the length of the reader enough to pass the length
check.

To fix that, now, the frame_size field holds the frame size excluding
the header size.

closes #434

### DIFF
--- a/hazelcast/connection.py
+++ b/hazelcast/connection.py
@@ -693,7 +693,8 @@ class _Reader(object):
         self._builder = builder
         self._bytes_read = 0
         self._bytes_written = 0
-        self._frame_size = 0
+        # Size of the frame excluding the header (SIZE_OF_FRAME_LENGTH_AND_FLAGS bytes)
+        self._frame_size = -1
         self._frame_flags = 0
         self._message = None
 
@@ -719,22 +720,21 @@ class _Reader(object):
                 return None
 
     def _read_frame(self):
-        n = self.length
-        if n < SIZE_OF_FRAME_LENGTH_AND_FLAGS:
-            # we don't have even the frame length and flags ready
-            return False
+        if self._frame_size == -1:
+            if self.length < SIZE_OF_FRAME_LENGTH_AND_FLAGS:
+                # we don't have even the frame length and flags ready
+                return False
 
-        if self._frame_size == 0:
             self._read_frame_size_and_flags()
 
-        if n < self._frame_size:
+        if self.length < self._frame_size:
             return False
 
         self._buf.seek(self._bytes_read)
-        size = self._frame_size - SIZE_OF_FRAME_LENGTH_AND_FLAGS
+        size = self._frame_size
         data = self._buf.read(size)
         self._bytes_read += size
-        self._frame_size = 0
+        self._frame_size = -1
         # No need to reset flags since it will be overwritten on the next read_frame_size_and_flags call
         frame = Frame(data, self._frame_flags)
         if not self._message:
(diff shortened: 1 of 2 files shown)


yuce (approved):
Looks good!

mdumandag:
Thanks for the review Yüce

PR #445, opened 2021-08-09
Follow-up aggregators implementation

mdumandag:
There were couple of missing things in the community PR for the
aggregators. This is an attempt to fix those. Namely,

- Some missing aggreagators are added (distinct, min_by, max_by).
  For them, some identified factories are also added, as they
  have different return values.
- Aggregator module is added to the API documentation
- A code sample is added
- Docstrings are updated
- Documentation is updated and corrected

### DIFF
--- a/docs/using_python_client_with_hazelcast_imdg.rst
+++ b/docs/using_python_client_with_hazelcast_imdg.rst
@@ -2086,17 +2086,20 @@ The ``aggregator`` module provides a wide variety of built-in aggregators. The
 full list is presented below:
 
 - ``count``
+- ``distinct``
 - ``double_avg``
 - ``double_sum``
+- ``fixed_point_sum``
 - ``floating_point_sum``
-- ``number_avg``
-- ``max_``
-- ``min_``
-- ``long_avg``
-- ``long_sum``
 - ``int_avg``
 - ``int_sum``
-- ``fixed_point_sum``
+- ``long_avg``
+- ``long_sum``
+- ``max_``
+- ``min_``
+- ``number_avg``
+- ``max_by``
+- ``max_by``
 
 These aggregators are used with the ``map.aggregate`` function, which takes an
 optional predicate argument.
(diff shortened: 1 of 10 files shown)


hazelcast/aggregator.py
```diff
@@ -89,245 +88,373 @@ def get_class_id(self):
...
-def long_avg(attribute_path=None):
-    """Creates long average aggregator.
+def max_(attribute_path=None):
+    """Creates an aggregator that calculates the max of the input values.
 
-    Does NOT accept ``None`` input values.
-    Accepts only long input values (primitive and boxed).
+    Accepts ``None`` input values and ``None`` extracted values.
+
+    Accepts any input value type that is comparable.
```
srknzl, line 375: We mentioned java types etc. in other docstrings. To make this `comparable` word more explicit, should we mention Comparable interface of Java? Same for min, max_by, min_by.

hazelcast/aggregator.py
```diff
@@ -89,245 +88,373 @@ def get_class_id(self):
...
+    Returns:
+        Aggregator[hazelcast.core.MapEntry]: An aggregator that calculates
+        the input value containing the maximum value extracted from the path.
+    """
+    return _MaxByAggregator(attribute_path)
+
+
+def min_by(attribute_path):
+    """Creates an aggregator that calculates the min of the input values
+    extracted from the given ``attribute_path`` and returns the input
```
srknzl, line 445: nit: None of these methods above contained `attribute_path` in the first paragraph of the docstring. Should we remove it from min_by and max_by too?

like this:

```
Creates an aggregator that calculates the min of the input values and ret...``
```

tests/integration/backward_compatible/proxy/map_test.py
```diff
@@ -790,32 +793,64 @@ def test_int_average_with_attribute_path(self):
...
+
+    def test_min_by(self):
+        min_item = self.map.aggregate(min_by("this"))
+        self.assertEqual("key-0", min_item.key)
+        self.assertEqual(0, min_item.value)
+
+    def test_min_by_with_predicate(self):
+        min_item = self.map.aggregate(min_by("this"), greater_or_equal("this", 10))
+        self.assertEqual("key-10", min_item.key)
+        self.assertEqual(10, min_item.value)
```
srknzl, line 853: We don't need to add test all of long, double, and int, that would be unnecessary. However, if we can test `distinct` with a map that includes same values, it would be better. Right now, the values are from 0..49, which are unique.

srknzl (approved):
LGTM, left some minor comments

docs/using_python_client_with_hazelcast_imdg.rst
```diff
@@ -2086,17 +2086,20 @@ The ``aggregator`` module provides a wide variety of built-in aggregators. The
...
 - ``int_avg``
 - ``int_sum``
-- ``fixed_point_sum``
+- ``long_avg``
+- ``long_sum``
+- ``max_``
+- ``min_``
+- ``number_avg``
+- ``max_by``
+- ``max_by``
```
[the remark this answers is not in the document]
mdumandag: No, I sorted them according to the class ids

hazelcast/aggregator.py
mdumandag: No, for them attribute path is optional, but for min_by and max_by it is required. So, I choose to document attribute_path behavior on the parameter docstring for others

hazelcast/aggregator.py
```diff
@@ -89,245 +88,373 @@ def get_class_id(self):
...
 
     Args:
-        attribute_path (str): extracts values from this path if given
+        attribute_path (str): Extracts values from this path, if given.
 
     Returns:
-        Aggregator[float]: an aggregator that calculates the average of
-        the input values
+        Aggregator[float]: An aggregator that calculates the sum of input the
+        values.
```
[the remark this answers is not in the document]
mdumandag: Fixed it as you suggested

hazelcast/aggregator.py
mdumandag: Improved the documentation for such aggregators. Please see the latest commit

tests/integration/backward_compatible/proxy/map_test.py
mdumandag: Improved the tests so that we have some duplicate values

srknzl: changes look good, thanks

PR #454, opened 2021-08-19
Projections follow up

mdumandag:
This is a follow up PR after merging the Projections PR. This is
a combination of lots of small things.

- Add documentation to properties of `MapEntry` so that they are
displayed
- Move map#project so that we maintain alphabetical order
- Add some corner case tests for projections and aggregations
- Used asserCountEqual on projection tests so that the tests will
be more durable, even if we add more items to map
- Add missing API documentation for projections
- Fix API documentation of projections around the return value documentations
- Add unit tests for the invalid projection inputs
- Make the projections code snippet simpler
- Add a code sample for projections

### DIFF
--- a/docs/using_python_client_with_hazelcast_imdg.rst
+++ b/docs/using_python_client_with_hazelcast_imdg.rst
@@ -2166,31 +2166,30 @@ See the following example.
     import hazelcast
 
     from hazelcast.core import HazelcastJsonValue
-    from hazelcast.predicate import greater_or_equal
+    from hazelcast.predicate import greater
     from hazelcast.projection import single_attribute, multi_attribute
 
     client = hazelcast.HazelcastClient()
     employees = client.get_map("employees").blocking()
 
-    employees.put(1, HazelcastJsonValue('{"Age": 23, "Height": 180, "Weight": 60}'))
-    employees.put(2, HazelcastJsonValue('{"Age": 21, "Height": 170, "Weight": 70}'))
+    employees.put(1, HazelcastJsonValue({"age": 25, "height": 180, "weight": 60}))
+    employees.put(2, HazelcastJsonValue({"age": 21, "height": 170, "weight": 70}))
+    employees.put(3, HazelcastJsonValue({"age": 40, "height": 175, "weight": 75}))
 
-    employee_ages = employees.project(single_attribute("Age"))
-    # Prints:
-    # The ages of employees are [21, 23]
-    print("The ages of employees are %s" % employee_ages)
+    ages = employees.project(single_attribute("age"))
 
-    # Run Single Attribute With Predicate
-    employee_ages = employees.project(single_attribute("Age"), greater_or_equal("Age", 23))
-    # Prints:
-    # The employee age is 23
-    print("The employee age is: %s" % employee_ages[0])
+    # Prints: "Ages of the employees are [21, 25, 40]"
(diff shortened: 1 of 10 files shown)


hazelcast/projection.py
```diff
@@ -48,7 +48,7 @@ def get_class_id(self):
 
 
 class _MultiAttributeProjection(_AbstractProjection):
-    def __init__(self, *attribute_paths):
+    def __init__(self, attribute_paths):
```
yuce, line 51: How about adding types to new classes and functions?

tests/integration/backward_compatible/proxy/map_test.py
```diff
@@ -796,6 +796,14 @@ def setUp(self):
     def tearDown(self):
         self.map.destroy()
 
+    def test_aggregate_with_none_aggregator(self):
+        with self.assertRaises(AssertionError):
+            self.map.aggregate(None)
```
yuce, line 801: Can be simplified as: `self.assertRaises(AssertionError, lambda: self.map.aggregate(None))` I think most of other assertRaises uses can also be put into a single line.

yuce (changes requested):
Looks good, but I think using `attribute` instead of `single_attribute` and `multi_attribute` would make our API simpler and more Pythonic.

tests/integration/backward_compatible/proxy/map_test.py
mdumandag: I personally prefer this style, and try to use it like this across the codebase. Advantages:

- You can easily put breakpoints
- You can put multiple lines without defining a closure

And IMHO, this one is more readable

docs/using_python_client_with_hazelcast_imdg.rst
```diff
@@ -2166,31 +2166,30 @@ See the following example.
...
-    employee_multi_attribute = employees.project(multi_attribute("Age", "Height"))
-    # Prints:
-    # Employee 1 age and height: [21, 170] Employee 2 age and height: [23, 180]
-    print("Employee 1 age and height: %s Employee 2 age and height: %s" % (employee_multi_attribute[0], employee_multi_attribute[1]))
+    filtered_ages = employees.project(single_attribute("age"), greater("age", 23))
+
+    # Prints: "Ages of the filtered employees are [25, 40]"
+    print("Ages of the filtered employees are %s" % filtered_ages)
+
+    attributes = employees.project(multi_attribute("age", "height"))
```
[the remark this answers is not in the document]
mdumandag: But they are two distinct things with distinct return types. One returns `list[any]`, the other `list[list[any]]` when used with project method. 

Also, I find such codes like the below a bit weird in the implementation. IMHO, a separation is better than this

```python
def attribute(*attrs):
    if len(attrs) == 1:
        return _SingleAttribute(attrs[0])
    return _MultiAttribute(attrs)
```

hazelcast/projection.py
mdumandag: I have added type hints to the projections module. I didn't type-hint the code that is coming from the superclass (IdentifiedDataSerializable), should I add type hints to those methods too (like get_class_id)? 

Also, I question. We have docstrings that define types for public methods, should we define type hints for them too? I have added type hints to them but I can remove them if you want

yuce (approved):

PR #455, opened 2021-08-20
Token providers can use the member address for token creation

yuce:


hazelcast/security.py
```diff
@@ -1,13 +1,17 @@
 from hazelcast.six import string_types
+from hazelcast.core import Address
 
 
 class TokenProvider(object):
     """TokenProvider is a base class for token providers."""
 
-    def token(self):
-        # type: (TokenProvider) -> bytes
+    def token(self, address):
```
mdumandag, line 8: Does it make sense to make this optional by providing `None` as the default value? In basic use cases, you don't need an address, you can simply ignore it.

yuce: Yeah, that'd be better: https://github.com/hazelcast/hazelcast-python-client/pull/455/commits/57f3d488c0cc7b3959134576e29c97f6eddde35d

tests/unit/token_provider_test.py
```diff
@@ -1,6 +1,7 @@
 import unittest
 
 from hazelcast.security import BasicTokenProvider
+from hazelcast.core import Address
```
mdumandag, line 4: Minor: Unused import

mdumandag (approved):

tests/unit/token_provider_test.py
yuce: Oops: https://github.com/hazelcast/hazelcast-python-client/pull/455/commits/d9ee08afd46e31f6cbbf8c9217a3e6ecddf418fc

mdumandag:
verify

yuce:
verify

PR #456, opened 2021-08-23
[API-700] v5 SQL: Initial PR

mdumandag:
This is an initial PR for the v5 SQL. It consists minimal changes
required to test the v5 SQL. Follow up PRs will come after that.

Here are the list of changes made:

- Set client version to 5.0
- Update remote controller scripts so that it uses hazelcast+hazelcast-sql
JARs instead of hazelcast-all, since hazelcast-all is not produced anymore.
Also, update some documentation about it.
- Use ints to represent years in DATE column type
- Use built-in types for DATE, TIME, TIMESTAMP, TIMESTAMP WITH TIMEZONE, and DECIMAL
which are datetime.date, datetime.time, datetime.datetime,
datetime.datetime(with non-None tzinfo), and decimal.Decimal.
- Update SQL tests and add some Jet tests
- Fix distributed objects tests by filtering internal objects out.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,9 +72,9 @@ author = u"Hazelcast Inc. Developers"
 # built documents.
 #
 # The short X.Y version.
-version = "4.2.1"
+version = "5.0"
 # The full version, including alpha/beta/rc tags.
-release = "4.2.1"
+release = "5.0"
 
 autodoc_member_order = "bysource"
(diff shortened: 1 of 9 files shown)


srknzl:
Will there be documentation and code samples PRs after this?

mdumandag:
@srknzl, yes I will send them in a separate PR

PR #457, opened 2021-08-24
[API-701] v5 SQL: Use different logic for getting query connections

mdumandag:
In v5, the logic behind selecting a query connection is changed
a little bit.

Formerly, we were relying on the load balancers to fetch the next
data member. Now, we are selecting a random connection from the
largest same-version (ignoring the patch version) member group.
This way, SQL will be available on the client-side during rolling
upgrade scenarios.


mdumandag:
verify

PR #469, opened 2021-09-08
Increase timeout used in `test_await_latch_with_timeout`

mdumandag:
Increased the timeout from 0.1 second to 1 second. Also, in Windows
due to low time resolution, we lowered the compared value slightly.

With increased timeout, comparing to a slightly lower value should
be OK since it shows that we are in fact waiting for the timeout.

Note that, the miscalculation comes from the fact that, the calculation
for the time passed in the Python side is precise to +-16ms. See
https://bugs.python.org/issue44328 and the Windows documentation for
the methods used by the CPython for more information.

### DIFF
--- a/tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
+++ b/tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
@@ -61,7 +61,7 @@ class CountDownLatchTest(CPTestCase):
 
         self.assertTrue(
             time_passed >= expected_time_passed,
-            "Time passed is less than %s, which is %s" % (timeout, expected_time_passed),
+            "Time passed is less than %s, which is %s" % (expected_time_passed, time_passed),
         )
 
     def test_await_latch_multiple_waiters(self):
--- a/tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
+++ b/tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
@@ -1,3 +1,4 @@
+import os
 from threading import Thread
 
 from hazelcast.errors import DistributedObjectDestroyedError, OperationTimeoutError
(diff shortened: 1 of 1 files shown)


tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
```diff
@@ -44,14 +45,23 @@ def test_await_latch_zero_timeout(self):
         self.assertFalse(latch.await_latch(0))
 
     def test_await_latch_with_timeout(self):
-        timeout = 0.1
+        timeout = 1
```
srknzl, line 48: Why we increased this? I mean, can you explain a little bit in detail why it will help? Also was this test failing in the past? Why and when it started to fail?

tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
```diff
@@ -44,14 +45,23 @@ def test_await_latch_zero_timeout(self):
...
+            # time passed being slightly less than the timeout. This is due to
+            # the low time resolution there (15-16ms). If we are on Windows, we
+            # lower our expectations and settle for a slightly lower value.
+            expected_time_passed *= 0.95
+
         self.assertTrue(
-            time_passed >= timeout,
-            "Time passed is less than %s, which is %s" % (timeout, time_passed),
+            time_passed >= expected_time_passed,
+            "Time passed is less than %s, which is %s" % (timeout, expected_time_passed),
```
srknzl, line 64: I think you renamed the wrong variable in this line

tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
mdumandag: It started failing once we started running tests on Windows. 

It will show that we are in fact waiting. 0.095 seconds might not show that we are waiting for timeout, so I made it 1 second to show that we are in fact waiting for the timeout. Note that, 0.95 for Windows also shows that we are waiting for quite long time

tests/integration/backward_compatible/proxy/cp/count_down_latch_test.py
mdumandag: Fixed in https://github.com/hazelcast/hazelcast-python-client/pull/469/commits/9f52471f47439ccdbe7cb30fc265b3c611e82cdf

srknzl (approved):

PR #472, opened 2021-09-14
[API-875] Deserialize `SqlRow`s lazily

### LINKED ISSUE #447: [TRACKING ISSUE] Lazy SQL OBJECT deserialization

See https://github.com/hazelcast/hazelcast/pull/19290 for details.

---

Fixes #19281

mdumandag:
This is a required change to be able to use parts of the row,
even if we cannot deserialize some column values.

Also, lazy deserialization is a plus on its own.

To accomplish this, similar to the Java client, I have changed the
code so that, we try to deserialize column values when we
call `get_object` or `get_object_with_index`. Note that,
just like the Java client

```
row.get_object("smth")
row.get_object("smth")
```

will result in two deserializations.
We could easily solve this with an extra object per row (something
like LazyDeserializable class with two fields: serialized and deserialized
where once we deserialize it, it assigns the result to deserialized and
return it for the subsequent calls). But, I didn't want to go that way due
to extra object cost.

Closes #447

### DIFF
--- a/hazelcast/sql.py
+++ b/hazelcast/sql.py
@@ -523,11 +523,12 @@ class SqlRowMetadata(object):
 class SqlRow(object):
     """One of the rows of an SQL query result."""
 
-    __slots__ = ("_row_metadata", "_row")
+    __slots__ = ("_row_metadata", "_row", "_deserialize_fn")
 
-    def __init__(self, row_metadata, row):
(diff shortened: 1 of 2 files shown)


zpqrtbnk (approved):
To the best of my Python knowledge this is OK with be and then we have to do the same in C# of course...

PR #475, opened 2021-09-15
[API-879] Wait for listener deregistration requests

### LINKED ISSUE #448: [TRACKING ISSUE] Remove listeners sync with the calling thread

See https://github.com/hazelcast/hazelcast/pull/19289 for details.

---

It turns out that there are some services relying on the removal
of a listener on the member when listener is removed from the client.
Example:
Cache listener count is kept per proxy. When listener is removed,
we decrease the listener count from the cache context.
Since listener removal is async, it could be the case that the cache
is destroyed from the calling thread before we decrement the value.
In such cases, we decrement but the incremented value is lost. Therefore,
we can end up with negative values.

The listener derregistration invoctions are not waited. It was
sync but we have changed the behavior as a side affect during
this fix https://github.com/hazelcast/hazelcast/pull/17646

We may solve this only for CacheListener but there could be more
services relying on sync listener removal. Therefore, as a fix,
we wait for invocation answer from the remote.

fixes https://github.com/hazelcast/hazelcast/issues/19269

mdumandag:
With this PR, we will wait for all the listener deregistration
requests to finish before returning `True` to the user.

It is mainly a precaution to the possible problems that might result
in services that expect sync removal of listeners.

While doing that, I have also dived into one of our technical debts
and corrected the implementation of the `combine_futures`.

It now waits for all futures to complete in all scenarios, as it should be,
and sets the result to the first exceptional completion, if some of
the input futures complete as such.

Closes #448

### DIFF
--- a/hazelcast/listener.py
+++ b/hazelcast/listener.py
@@ -3,7 +3,6 @@ import sys
 import threading
 from uuid import uuid4
 
-import hazelcast.future
 from hazelcast import six
 from hazelcast.errors import HazelcastError, HazelcastClientNotActiveError, TargetDisconnectedError
 from hazelcast.future import combine_futures, ImmediateFuture
@@ -118,9 +117,10 @@ class ListenerService(object):
                         return
 
                     _logger.warning(
-                        "Deregistration of listener with ID %s has failed for address %s",
+                        "Deregistration of listener with ID %s has failed for address %s: %s",
                         user_registration_id,
                         captured_connection.remote_address,
+                        e,
                     )
 
                 invocation.future.add_done_callback(handler)
--- a/hazelcast/listener.py
+++ b/hazelcast/listener.py
@@ -3,6 +3,7 @@ import sys
 import threading
 from uuid import uuid4
 
+import hazelcast.future
 from hazelcast import six
 from hazelcast.errors import HazelcastError, HazelcastClientNotActiveError, TargetDisconnectedError
 from hazelcast.future import combine_futures, ImmediateFuture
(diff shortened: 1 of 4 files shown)


sancar (approved):

hazelcast/listener.py
```diff
@@ -105,24 +107,27 @@ def deregister_listener(self, user_registration_id):
...
+                    if isinstance(
+                        e, (HazelcastClientNotActiveError, IOError, TargetDisconnectedError)
+                    ):
+                        return
+
+                    _logger.warning(
+                        "Deregistration of listener with ID %s has failed for address %s",
+                        user_registration_id,
+                        captured_connection.remote_address,
+                    )
```
yemreinci, line 124: Java client also [logs the exception](https://github.com/hazelcast/hazelcast/blob/b6c6c270ddffb20ce7dec1954e1d75ac4e5866ab/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/listener/ClientListenerServiceImpl.java#L343) here.

yemreinci (approved):
`combine_futures` looks perfect :heavy_check_mark: 

`deregister_listener` seems to follow Java except a small difference. 

And I have a small question: We pop from the local registration dictionary `_active_registrations` before receiving a successful deregistration response from the members. So, can there be an erroneous case where the member still thinks we have a listener registered and keep sending events?

hazelcast/listener.py
mdumandag: Thanks. I added the error message to the log message. Also, removed the unused import that I added erronously.

mdumandag:
@yemreinci Yes, correct. I think we discussed this a while ago during one of our standups, discussed various solutions for that problem, and reached this state. 

The client always removes its local listener registration, regardless of the listener deregistration request outcome. When a member sends an event that we don't know, we will simply ignore it, but as you said it will keep coming to the client

PR #478, opened 2021-09-17
Make `SqlRow` support `[]` operator

mdumandag:
This is mainly an enhancement for easier access for columns in the row.
Now, the `SqlRow` implements the `__getitem__` magic method.

Implementation of the method simply performs a type check and calls
`get_object_with_index` if the type of the passed parameter is an
integer type. For any other type, it calls `get_object`. Hence,
it will throw for any other type than `int` or `str`.

Also, put a piece of information about this new way to the documentation
& code samples.

### DIFF
--- a/examples/sql/sql_example.py
+++ b/examples/sql/sql_example.py
@@ -67,6 +67,11 @@ with client.sql.execute("SELECT * FROM customers") as result:
         # Get the object with the column index
         is_active = row.get_object_with_index(is_active_index)
 
+        # Rows can also be used similar to lists
+        # with integer column indexes
+
+        # is_active = row[is_active_index]
+
         print(name, age, is_active)
 
 # Construct a statement object to control the properties of the query
@@ -86,13 +91,20 @@ with client.sql.execute_statement(statement) as result:
         key = row.get_object("__key")
         age = row.get_object("age")
 
+        # Rows can also be used similar to dictionaries
+        # with string column names
+
+        # key = row["__key"]
+        # age = row["age"]
+
         print(key, age)
 
 # Parameters can be passed directly in the basic execution syntax
 result = client.sql.execute("SELECT this FROM customers WHERE age > ? AND age < ?", 30, 40)
 
 for row in result:
-    customer = row.get_object("this")
+    # Access columns with [] operator
+    customer = row["this"]
     print(customer)
 
 # Query can be closed explicitly
--- a/hazelcast/sql.py
+++ b/hazelcast/sql.py
@@ -521,7 +521,24 @@ class SqlRowMetadata(object):
 
 
 class SqlRow(object):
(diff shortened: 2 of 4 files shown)


examples/sql/sql_example.py
```diff
@@ -86,13 +91,20 @@ def __repr__(self):
         key = row.get_object("__key")
         age = row.get_object("age")
 
+        # Rows can also be used similar to dictionaries
+        # with string column names
+
+        # key = row["__key"]
+        # age = row["age"]
+
```
srknzl, line 99: I think we can move this to line 48 which is the first get_object() call, it can be overlooked here. Optionally, we can write this comment multiple times

There is another a couple of get_object calls at line 58 and 59

hazelcast/sql.py
```diff
@@ -609,6 +626,12 @@ def metadata(self):
         """SqlRowMetadata: The row metadata."""
         return self._row_metadata
 
+    def __getitem__(self, item):
+        if isinstance(item, six.integer_types):
+            return self.get_object_with_index(item)
+
+        return self.get_object(item)
```
srknzl, line 633: Just to conform to what python docs says: 


> Called to implement evaluation of self[key]. For sequence types, the accepted keys should be integers and slice objects. Note that the special interpretation of negative indexes (if the class wishes to emulate a sequence type) is up to the __getitem__() method. If key is of an inappropriate type, TypeError may be raised; if of a value outside the set of indexes for the sequence (after any special interpretation of negative values), IndexError should be raised. For mapping types, if key is missing (not in the container), KeyError should be raised.


at
https://docs.python.org/3/reference/datamodel.html#object.__getitem__

Also note that we raise `ValueError` if key is a string and there is no column with that name.  According to documentation we should raise a `KeyError`.

The following conforms to the documentation:

```suggestion
        if isinstance(item, six.integer_types):
            return self.get_object_with_index(item)

        if isinstance(item, six.string_types):
            index = self._row_metadata.find_column(column_name)
            if index == SqlRowMetadata.COLUMN_NOT_FOUND:
                raise KeyError("Column '%s' doesn't exist" % column_name)
            return self.get_object(item)

        raise TypeError
```

srknzl (approved):
My comments are minor so this is an "approve" from me

examples/sql/sql_example.py
mdumandag: I didn't want to put these comments everywhere and this code sample is meant to be read from beginning to end to understand different ways to use SQL service. So, I think it it okay to leave it like this

hazelcast/sql.py
mdumandag: I feel like it is okay to raise `ValueError` here. As our documentation suggests, this is just an easier way to use `get_object` and `get_object_with_index`.

PR #479, opened 2021-09-17
Update SqlService API documentation

mdumandag:
- Removed the warning about the BETA status
- Similar to Java client, simplified the API doc as most of the
things we mention here is already mentioned in the actual
SQL documentation
- Add a warning about enabling Jet and adding `hazelcast-sql`
to the member's classpath
- And other minor updates.

### DIFF
--- a/hazelcast/sql.py
+++ b/hazelcast/sql.py
@@ -23,86 +23,39 @@ _logger = logging.getLogger(__name__)
 class SqlService(object):
     """A service to execute SQL statements.
 
-    The service allows you to query data stored in a
-    :class:`Map <hazelcast.proxy.map.Map>`.
-
     Warnings:
 
-        The service is in beta state. Behavior and API might change
-        in future releases.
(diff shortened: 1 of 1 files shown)


hazelcast/sql.py
```diff
@@ -23,86 +23,39 @@
...
-    key and a value and then merges them into a single column set.
-    In case the key and the value have columns with the same name, the
-    key takes precedence.
+        In order to use this service, Jet engine must be enabled on the members
+        and the ``hazelcast-sql`` module must be in the classpath of the
+        members.
 
-    Columns are extracted from objects as follows (which happens on the
-    server-side):
+        If you are using the our CLI, Docker image, or distributions to start
```
srknzl, line 32:
```suggestion
        If you are using the CLI, Docker image, or distributions to start
```

srknzl (approved):

hazelcast/sql.py
srknzl: or "If you are using our CLI

PR #480, opened 2021-09-20
[API-900] Wait for the initial response in the SQL queries

mdumandag:
Up until now, we were returning the `SqlResult` immediately after
initiating the query to the server, without waiting for the first
response.

The server might not be able to handle some consequences of such
an API. To align the behavior with the Java implementation and
make it more intuitive, we now wait for the initial response
in the `execute` and `execute_statement` methods.

With that change, the return types are changed from `SqlResult` to
`Future[SqlResult]`. Also, I got rid of the Futures in the return types
of the methods over the `SqlResult` except for the `close`, as the
`close` might in fact go to the server.

With this change, I cleaned up the implementation so that it reflects
the new behavior better.

Also, we now return a rejected Future immediately if the first execute
the request fails.

### DIFF
--- a/docs/using_python_client_with_hazelcast_imdg.rst
+++ b/docs/using_python_client_with_hazelcast_imdg.rst
@@ -1563,7 +1563,7 @@ The following code prints names of the employees whose age is less than 30:
 
 .. code:: python
 
-    result = client.sql.execute("SELECT name FROM emp WHERE age < ?", 30)
+    result = client.sql.execute("SELECT name FROM emp WHERE age < ?", 30).result()
 
     for row in result:
         name = row.get_object("name")
@@ -1658,6 +1658,7 @@ it:
 .. code:: python
 
     client = hazelcast.HazelcastClient()
+    employees = client.get_map("employees").blocking()
 
     mapping_query = """
     CREATE MAPPING employees (
@@ -1671,17 +1672,14 @@ it:
     );
     """
 
-    with client.sql.execute(mapping_query) as result:
-        # wait until query completes
-        result.update_count().result()
+    client.sql.execute(mapping_query).result():
 
-    employees = client.get_map("employees").blocking()
     employees.set(1, HazelcastJsonValue({"name": "John Doe", "salary": 60000}))
     employees.set(2, HazelcastJsonValue({"name": "Jane Doe", "salary": 80000}))
 
     select_query = "SELECT name FROM employees WHERE salary > 75000"
 
-    with client.sql.execute(select_query) as result:
+    with client.sql.execute(select_query).result() as result:
         for row in result:
             name = row.get_object("name")
             # or, you can use the [] operator
--- a/docs/using_python_client_with_hazelcast_imdg.rst
+++ b/docs/using_python_client_with_hazelcast_imdg.rst
@@ -1672,7 +1672,7 @@ it:
     );
     """
 
-    client.sql.execute(mapping_query).result():
+    client.sql.execute(mapping_query).result()
 
     employees.set(1, HazelcastJsonValue({"name": "John Doe", "salary": 60000}))
     employees.set(2, HazelcastJsonValue({"name": "Jane Doe", "salary": 80000}))
--- a/hazelcast/sql.py
+++ b/hazelcast/sql.py
@@ -3,7 +3,7 @@ import uuid
 from threading import RLock
 
 from hazelcast.errors import HazelcastError
-from hazelcast.future import Future, ImmediateFuture
+from hazelcast.future import Future, ImmediateFuture, ImmediateExceptionFuture
 from hazelcast.invocation import Invocation
 from hazelcast.util import (
     UUIDUtil,
@@ -589,7 +589,7 @@ class SqlRow(object):
 class _ExecuteResponse(object):
     """Represent the response of the first execute request."""
 
-    __slots__ = ("row_metadata", "row_page", "update_count", "error")
+    __slots__ = ("row_metadata", "row_page", "update_count")
 
     def __init__(self, row_metadata, row_page, update_count, error):
         self.row_metadata = row_metadata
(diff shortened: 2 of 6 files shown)


tests/integration/backward_compatible/sql_test.py
```diff
@@ -178,20 +217,20 @@ def test_execute_with_params(self):
...
-        result = self.client.sql.execute(
-            'SELECT * FROM "%s" WHERE __key > ? AND this > ?' % self.map_name, 5
-        )
 
         with self.assertRaises(HazelcastSqlError):
+            result = self.execute(
+                'SELECT * FROM "%s" WHERE __key > ? AND this > ?' % self.map_name, 5
+            )
             for _ in result:
                 pass
```
srknzl, line 226:
```suggestion
        with self.assertRaises(HazelcastSqlError):
            self.execute(
                'SELECT * FROM "%s" WHERE __key > ? AND this > ?' % self.map_name, 5
            )
```

tests/integration/backward_compatible/sql_test.py
```diff
@@ -178,20 +217,20 @@ def test_execute_with_params(self):
...
 
     def test_execute_with_mismatched_params_when_params_has_more(self):
         self._create_mapping()
         self._populate_map()
-        result = self.client.sql.execute('SELECT * FROM "%s" WHERE this > ?' % self.map_name, 5, 6)
 
         with self.assertRaises(HazelcastSqlError):
+            result = self.execute('SELECT * FROM "%s" WHERE this > ?' % self.map_name, 5, 6)
             for _ in result:
                 pass
```
srknzl, line 235:
```suggestion
            self.execute('SELECT * FROM "%s" WHERE this > ?' % self.map_name, 5, 6)
```

tests/integration/backward_compatible/sql_test.py
```diff
@@ -236,9 +275,9 @@ def test_execute_statement_with_mismatched_params_when_params_has_more(self):
         self._populate_map()
         statement = SqlStatement('SELECT * FROM "%s" WHERE this > ?' % self.map_name)
         statement.parameters = [5, 6]
-        result = self.client.sql.execute_statement(statement)
 
         with self.assertRaises(HazelcastSqlError):
+            result = self.execute_statement(statement)
             for _ in result:
                 pass
```
srknzl, line 282:
```suggestion
            self.execute_statement(statement)
```

tests/integration/backward_compatible/sql_test.py
```diff
@@ -279,11 +318,11 @@ def test_execute_statement_with_copy(self):
...
 
-        result = self.client.sql.execute_statement(copy_statement)
+        result = self.execute_statement(copy_statement)
         self.assertEqual([9], [row.get_object_with_index(0) for row in result])
 
-        result = self.client.sql.execute_statement(statement)
         with self.assertRaises(HazelcastSqlError):
+            result = self.execute_statement(statement)
             for _ in result:
                 pass
```
srknzl, line 327:
```suggestion
            self.execute_statement(statement)
```

tests/integration/backward_compatible/sql_test.py
```diff
@@ -308,9 +347,9 @@ def test_execute_statement_with_expected_result_type_of_update_count_when_rows_a
         self._populate_map()
         statement = SqlStatement('SELECT * FROM "%s"' % self.map_name)
         statement.expected_result_type = SqlExpectedResultType.UPDATE_COUNT
-        result = self.client.sql.execute_statement(statement)
 
         with self.assertRaises(HazelcastSqlError):
+            result = self.execute_statement(statement)
             for _ in result:
                 pass
```
srknzl, line 354:
```suggestion
            self.execute_statement(statement)
```

docs/using_python_client_with_hazelcast_imdg.rst
```diff
@@ -1671,17 +1672,14 @@ it:
     );
     """
 
-    with client.sql.execute(mapping_query) as result:
-        # wait until query completes
-        result.update_count().result()
+    client.sql.execute(mapping_query).result():
```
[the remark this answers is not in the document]
mdumandag: This is already changed in the #482. So, lets leave it like this

hazelcast/sql.py
```diff
@@ -848,13 +847,13 @@ def on_next_row(row_future):
     To get the number of rows updated by the query, use the
     :func:`update_count`. ::
 
-        update_count = client.sql.execute("SELECT ...").update_count().result()
+        update_count = client.sql.execute("UPDATE ...").result().update_count()
```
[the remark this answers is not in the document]
mdumandag: I don't know but I copied this from the Java side. I assume they did it like this because currently, we don't have any queries that return update count

hazelcast/sql.py
```diff
@@ -1159,109 +1124,31 @@ def _on_fetch_response(self, page):
...
-                self._execute_response.set_result(response)
-            else:
-                # Result only contains the update count.
-                response = _ExecuteResponse(None, None, update_count)
-                self._execute_response.set_result(response)
-
-                # There is nothing more we can get from the server.
-                self._closed = True
+        return (
+            execute_response.row_page is None  # Just an update count
```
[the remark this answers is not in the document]
mdumandag: This is the same and more inline with the check below

tests/integration/backward_compatible/sql_test.py
mdumandag: Can't do that, for backward compatibility tests as 4.x clients throw while using SQL result

hazelcast/sql.py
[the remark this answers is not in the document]
mdumandag: After a second look, the row page might be null, let me change that

srknzl (approved):

PR #481, opened 2021-09-20
Add missing `remove_interceptor` API to the Map

mdumandag:
Also, added a test for the `add_interceptor` method.

Protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/399

### DIFF
--- a/tests/integration/backward_compatible/proxy/map_test.py
+++ b/tests/integration/backward_compatible/proxy/map_test.py
@@ -81,6 +81,27 @@ class EntryProcessor(IdentifiedDataSerializable):
         return self.CLASS_ID
 
 
+class MapGetInterceptor(IdentifiedDataSerializable):
+
+    FACTORY_ID = 666
+    CLASS_ID = 6
+
+    def __init__(self, prefix):
+        self.prefix = prefix
+
+    def write_data(self, object_data_output):
+        write_string_to_output(object_data_output, self.prefix)
+
+    def read_data(self, object_data_input):
+        pass
+
+    def get_factory_id(self):
+        return self.FACTORY_ID
+
+    def get_class_id(self):
+        return self.CLASS_ID
+
+
 class MapTest(SingleMemberTestCase):
     @classmethod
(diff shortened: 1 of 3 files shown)


tests/integration/backward_compatible/proxy/map_test.py
```diff
@@ -596,6 +617,31 @@ def test_values_with_predicate(self):
     def test_str(self):
         self.assertTrue(str(self.map).startswith("Map"))
 
+    def test_add_interceptor(self):
+        interceptor = MapGetInterceptor("hell")
```
yuce, line 621: I don't know why 666 was used as the factory ID in the test jar, but we probably can rename this better :D

tests/integration/backward_compatible/proxy/map_test.py
```diff
@@ -596,6 +617,31 @@ def test_values_with_predicate(self):
...
+        self.assertIsNotNone(registration_id)
+
+        self.map.set(1, "o")
+        value = self.map.get(1)
+        self.assertEqual("hello", value)
+
+    def test_remove_interceptor(self):
+        skip_if_client_version_older_than(self, "5.0")
+
+        interceptor = MapGetInterceptor("hell")
```
yuce, line 632: It would be better to use a less-provocative prefix :)

yuce (approved):
Just two trivial comments.

tests/integration/backward_compatible/proxy/map_test.py
yuce: How about `hello`?

mdumandag: Yeah, probably :smile:

tests/integration/backward_compatible/proxy/map_test.py
mdumandag: This was the first thing that came to my mind when I saw 666 :smiling_imp: 

Anyways, I changed the texts to be nicer

PR #482, opened 2021-09-21
[API-899] Update SQL documentation according to recent changes

mdumandag:
This PR updates the SQL documentation at the client-side with
the recent changes done to the reference manual.

We mostly follow a similar structure, but we go into more detail
for Portable and JSON queries in these documents as these are
the most intuitive ways to run SQL queries from the Python client.

Note that, the code samples in this PR are written according to
recent PRs, some of which not merged yet.

### DIFF
--- a/docs/using_python_client_with_hazelcast_imdg.rst
+++ b/docs/using_python_client_with_hazelcast_imdg.rst
@@ -1349,8 +1349,10 @@ the entry processor in the ``Map`` methods. See the following example.
 SQL
 ---
 
-You can use SQL to query map entries in Hazelcast as well as real-time event
-streams and data at rest in other data stores.
+You can use SQL to query data in maps, Kafka topics, or a variety of file
+systems. Results can be sent directly to the client or inserted into maps or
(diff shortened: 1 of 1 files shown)


docs/using_python_client_with_hazelcast_imdg.rst
```diff
@@ -1349,193 +1349,95 @@ the entry processor in the ``Map`` methods. See the following example.
 SQL
 ---
```
srknzl, line 1351: maybe we can add a section corresponding to https://docs.hazelcast.com/hazelcast/5.0/sql/improving-performance

srknzl (approved):

docs/using_python_client_with_hazelcast_imdg.rst
mdumandag: OK, added that section. But 

```
To find out more about making sure certain map entries are stored on the same member, see Data Affinity.
```

part is not available in the Python client yet so I removed that part.

PR #495, opened 2021-10-06
Update GA workflow files for the 4.2.z branch

mdumandag:
While sending a PR to the 4.2.z branch, I saw that it uses the GA
workflows defined in that branch, so I am updating the test runner
and jira trigger workflows so that, the test runner creates
jobs that would pass the merge checks.

The only difference between this test runner and the one in the
master is that, this one does not check broken links in the readme.


srknzl (approved):
lgtm but won't we need to update docs and check links in them? we may do so in a patch release.

srknzl:
note that documentation generation checks for all dead links, not just in README (as far as I know)

mdumandag:
@srknzl I will do that in a later PR, I just wanted the merge the prerequisite for all others

PR #510, opened 2021-11-22
Drop support for Python 2.7, 3.4, and 3.5

### LINKED ISSUE #350: Drop Python 2.7 support
Continuing to maintain backward-compatible support for python2.7 greatly increases the maintenance burden on contributors, scares off new contributors, forces additional dependencies, and prevents (or slows) the adoption of new features.

https://python3statement.org/

I would also recommend dropping support for python `3.5`, but IMO dropping `2.7` is a bigger deal.

mdumandag:
This PR drops support for Python 2.7, 3.4, and 3.5 and adds a matrix
item to test the client code on Python 3.10.

Those aforementioned versions are removed from/added to the corresponding
workflow matrix.

I also modified the PR runner so that it runs tests on Python 3.6 and 3.10.

Also, I have removed now unnecessary Python version checks from the
requirements-*.txt files and modified the setup.py to reflect which
Python versions we now support.

Another change is done to the test runner, as we replaced `nose` with
`pytest`, as `nose` is not compatible with Python 3.10.

Closes #350


yuce (approved):
Thanks, looks good!

utku-caglayan (approved):
lgtm!

PR #512, opened 2021-12-07
Try to run enterprise tests on PR builders

mdumandag:
This is a PR to test whether we can run enterprise tests on PR builders
if we define the secrets used in the runners in our fork.


mdumandag:
It didn't work :(

Even though I defined the `HAZELCAST_ENTERPRISE_KEY` secret in my fork, I wasn't able to use it in the workflow.

PR #516, opened 2021-12-10
Remove useless object inheritance

mdumandag:
We were using `object` inheritance to use new-style classes in Python 2.
However, that is not necessary anymore, as we have dropped support for
Python 2.

This PR simply replaces all occurrences of `(object):` string with
`:`, to get rid of that useless base object class.

Protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/406


yuce (approved):
Looks good!

mdumandag:
It is a pretty simple change, merging it with one approval to not get any merge conflicts for the upcoming changes.

PR #517, opened 2021-12-13
test_runner.yaml is updated with event type pull_request_target

alisengul53:
This PR is changing the event trigger mechanism for PRs. `pull_request` event type is changed to `pull_request_target`
`pull_request_target` is a little bit dangerous because even it is triggered by a PR, it runs the base branch code and also it can reach the repository secrets

https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

For example, normal checkout action checks out the base branch's code and the running workflow is the base branch's workflow instead of the PR.

Another example, if you make some changes on workflow and create a PR, these changes won't affect the triggered workflow for the PR. The triggered workflow will be the base branch's workflow

That is why we need to be careful about this event type.

In order to be safe, a label check is added to workflow. It is triggered when it is labeled but jobs are run if it is labeled as `pr_builder`. So if there is another PR coming from community, we can check changes and then add the label to trigger.
I set a label name but we can also run jobs when it is labeled without checking label's name.

### DIFF
--- a/.github/workflows/test_runner.yml
+++ b/.github/workflows/test_runner.yml
@@ -4,13 +4,15 @@ on:
     branches:
       - master
       - '[45].*.z'
-  pull_request:
+  pull_request_target:
+    types: [labeled]
     branches:
       - master
       - '[45].*.z'
 jobs:
   run-tests:
     runs-on: ${{ matrix.os }}
+    if: contains(github.event.pull_request.labels.*.name, 'pr_builder') || github.event_name == 'push'
     name: Run tests with Python ${{ matrix.python-version }} on ${{ matrix.os }}
     strategy:
       matrix:
@@ -27,8 +29,17 @@ jobs:
         with:
           distribution: 'adopt'
           java-version: '8'
-      - name: Checkout to code
+          
+      - name: Checkout code for push event
+        if: github.event_name == 'push'
         uses: actions/checkout@v2
+
+      - name: Checkout code for PR
+        if: github.event_name == 'pull_request_target'
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          
       - name: Install dependencies
         run: |
           pip install -r requirements-test.txt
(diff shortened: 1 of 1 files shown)


yemreinci (approved):
Looks safe to me. It would be much better if there was a way to make it trigger without a label for PRs coming from our organization.

Also, maybe we can find similar repos on Github that does this? It looks OK to me but I'm still a bit hesitant as there might still be something we miss.

ldziedziul:
So as a hacker I have to push an innocent PR first, wait for a `pr_builder` label and then push a secret-stealing commit, right?

ldziedziul:
You could also consider enforcing approval for all outside contributors: https://docs.github.com/en/enterprise-cloud@latest/admin/policies/enforcing-policies-for-your-enterprise/enforcing-policies-for-github-actions-in-your-enterprise#enforcing-a-policy-for-approval-of-pull-requests-from-outside-collaborators

alisengul53:
> So as a hacker I have to push an innocent PR first, wait for a `pr_builder` label and then push a secret-stealing commit, right?

Yes as a hacker you can try that but it won't work. Because the workflow is checks out the code when it is triggered and it is triggered when the label is added. If you push a new commit, the workflow is not triggered again. A member of organization should remove and add the label again.

alisengul53:
> You could also consider enforcing approval for all outside contributors: https://docs.github.com/en/enterprise-cloud@latest/admin/policies/enforcing-policies-for-your-enterprise/enforcing-policies-for-github-actions-in-your-enterprise#enforcing-a-policy-for-approval-of-pull-requests-from-outside-collaborators

It is eligible for `pull_request` event trigger but not for `pull_request_target`
>Note: Workflows triggered by pull_request_target events are run in the context of the base branch. Since the base branch is considered trusted, workflows triggered by these events will always run, regardless of approval settings.

alisengul53:
Will be handled with another PR

PR #519, opened 2021-12-23
[API-1048] Make intellisense and type hints available on blocking proxies

mdumandag:
This PR attempts to provide type hints and intellisense for the
blocking proxies by introducing blocking versions of the proxies
that implement the same interface with different return types.

With this approach, we don't have to duplicate the documentation,
and return the blocking subclasses directly to the user.

Actually, the blocking proxies are not really subclasses, but very
thin wrappers around the public APIs of the proxies. They look like
subclasses to make the documentation of the actual proxies available
to them.

This makes mypy complain, but it is possible to ignore such
warnings about return type mismatches between child and parent classes.
So, we could work around that problem once we start using mypy.

Also, we have modified the return types and documentation of transactional
proxies. It wasn't possible to use them in non-blocking mode so I changed
the return types and documentation to reflect that.


hazelcast/cluster.py
```diff
@@ -62,49 +63,63 @@ class ClusterService:
...
+        member_removed: typing.Callable[[MemberInfo], None] = None,
+        fire_for_existing=False,
+    ) -> str:
         """
         Adds a membership listener to listen for membership updates.
 
-        It will be notified when a member is added to cluster or removed from cluster.
-        There is no check for duplicate registrations, so if you register the listener
-        twice, it will get events twice.
+        It will be notified when a member is added to cluster or removed from
```
yuce, line 75: How about: "... added to the cluster or removed from the cluster ..."?

hazelcast/cluster.py
```diff
@@ -31,10 +32,10 @@ class ClientInfo:
     __slots__ = ("uuid", "address", "name", "labels")
 
     def __init__(self, client_uuid, address, name, labels):
-        self.uuid = client_uuid
-        self.address = address
-        self.name = name
-        self.labels = labels
+        self.uuid: uuid.UUID = client_uuid
```
yuce, line 35: How about annotating the `client_uuid` argument instead of `self.uuid`?

hazelcast/core.py
```diff
@@ -412,23 +415,23 @@ class HazelcastJsonValue:
     None values are not allowed.
     """
 
-    def __init__(self, value):
+    def __init__(self, value: typing.Union[str, typing.Any]):
```
yuce, line 418: Can `typing.Union[str, typing.Any]` be simplified to `typing.Any`? or just leave out the annotation altogether?

hazelcast/future.py
```diff
@@ -100,13 +102,13 @@ def exception(self):
...
 
-    def traceback(self):
+    def traceback(self) -> typing.Optional[types.TracebackType]:
         """Traceback of the exception."""
         self._reactor_check()
         self._event.wait()
         return self._traceback
 
-    def add_done_callback(self, callback):
+    def add_done_callback(self, callback: typing.Callable[["Future"], None]) -> None:
```
yuce, line 111: Does it make sense to move `typing.Callable["Future"]` to `hazelcast.typing`?

hazelcast/future.py
```diff
@@ -127,14 +129,19 @@ def _invoke_cb(self, callback):
         except:
             _logger.exception("Exception when invoking callback")
 
-    def continue_with(self, continuation_func, *args):
+    def continue_with(
+        self,
+        continuation_func: typing.Callable[..., typing.Union["Future", typing.Any]],
```
yuce, line 134: Can `typing.Union["Future", typing.Any]` be simplified to `typing.Any`?

hazelcast/predicate.py
```diff
@@ -360,9 +362,10 @@ def __init__(self, predicate, page_size, comparator=None):
         self._internal_predicate = predicate
         self._page_size = page_size
         self._page = 0  # initialized to be on first page
-        self.comparator = comparator
-        self.iteration_type = IterationType.ENTRY
-        self.anchor_list = []  # List of pairs: (nearest page, (anchor key, anchor value))
+        self.comparator: typing.Any = comparator
+        self.iteration_type: int = IterationType.ENTRY
+        # List of pairs: (nearest page, (anchor key, anchor value))
+        self.anchor_list: typing.List[typing.Tuple[int, typing.Tuple[typing.Any, typing.Any]]] = []
```
yuce, line 368: `typing.Tuple[typing.Any, typing.Any]` is equivalent to `tuple` or `typing.Tuple`, you may consider using the shorter alternatives.

hazelcast/predicate.py
```diff
@@ -586,7 +589,7 @@ def not_(predicate):
     return _NotPredicate(predicate)
 
 
-def between(attribute, from_, to):
+def between(attribute: str, from_: typing.Any, to: typing.Any) -> Predicate:
```
yuce, line 592: I think `typing.Any`s can be safely left out. There are more throughout the code.

hazelcast/projection.py
```diff
@@ -49,9 +51,8 @@ def get_class_id(self):
         return 0
 
 
-class _MultiAttributeProjection(_AbstractProjection):
-    def __init__(self, attribute_paths):
-        # type: (list[str]) -> None
+class _MultiAttributeProjection(_AbstractProjection[typing.List[typing.Any]]):
```
yuce, line 54: `typing.List[typing.Any]` can be simplified to `typing.List` or just `list`. There are more.

yuce (commented):
Here are a few ideas. Continuing to review.

hazelcast/proxy/cp/fenced_lock.py
```diff
@@ -117,15 +119,15 @@ def lock(self):
         invocation_uuid = uuid.uuid4()
         return self._do_lock(current_thread_id, invocation_uuid)
 
-    def try_lock(self, timeout=0):
+    def try_lock(self, timeout: Numeric = 0) -> Future[int]:
```
yuce, line 122: I think `timeout` should be just `float`.

hazelcast/proxy/flake_id_generator.py
```diff
@@ -10,26 +10,30 @@
...
-    the useful lifespan of the generator to little less than 70 years (until ~2088). The sequence component
-    is 6 bits. If more than 64 IDs are requested in single millisecond, IDs will gracefully overflow to the next
-    millisecond and uniqueness is guaranteed in this case. The implementation does not allow overflowing
-    by more than 15 seconds, if IDs are requested at higher rate, the call will block. Note, however, that
-    clients are able to generate even faster because each call goes to a different (random) member and
-    the 64 IDs/ms limit is for single member.
+    """A cluster-wide unique ID generator. Generated IDs are int values and
+    are k-ordered (roughly ordered). IDs are in the range from 0 to 2^63 - 1.
+
+    The IDs contain timestamp component and a node ID component, which is
```
yuce, line 16: "... contain a timestamp ..."

hazelcast/proxy/flake_id_generator.py
```diff
@@ -46,24 +50,28 @@ def __init__(self, service_name, name, context):
...
         )
 
-    def new_id(self):
+    def new_id(self) -> Future[int]:
         """Generates and returns a cluster-wide unique ID.
 
-        This method goes to a random member and gets a batch of IDs, which will then be returned
-        locally for limited time. The pre-fetch size and the validity time can be configured.
+        This method goes to a random member and gets a batch of IDs, which will
+        then be returned locally for limited time. The pre-fetch size and the
```
yuce, line 57: "... for a limited time."

hazelcast/predicate.py
mdumandag: I did this on purpose, to make it more clear that it contains two elements of type any. If it is not a problem for you, I would like to keep it like that.

I also used typing.Any or None return type in places to make sure that I did not miss anything.

mdumandag: BTW, we can't make use of `tuple` or `list` as they are not available in Python 3.6

hazelcast/predicate.py
mdumandag: See my comment above. It is done on purpose

hazelcast/proxy/cp/atomic_long.py
```diff
@@ -52,116 +55,116 @@ def compare_and_set(self, expect, update):
...
         Notes:
-            ``function`` must be an instance of ``IdentifiedDataSerializable`` or
-            ``Portable`` that has a counterpart that implements the
-            ``com.hazelcast.core.IFunction`` interface registered on the server-side with
-            the actual implementation of the function to be applied.
+            ``function`` must be an instance Hazelcast serializable type that
+            has a counterpart that implements the
+            ``com.hazelcast.core.IFunction`` interface registered on the
+            server-side with the actual implementation of the function to be
+            applied.
```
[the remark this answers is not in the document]
mdumandag: I will change it to 
```
            ``function`` must be an instance of Hazelcast serializable type.
            It must have a counterpart registered in the server-side that
            implements the ``com.hazelcast.core.IFunction`` interface with
            the actual logic of the function to be applied.
```

WDYT?

hazelcast/proxy/cp/atomic_reference.py
```diff
@@ -6,35 +9,40 @@
...
-      deserialized on the other side., e.g., because a method like `alter()` is executed.
-      This deserialization is done for every call that needs to have the object instead
-      of the binary content, so be careful with expensive object graphs that need to be
-      deserialized.
-    - If you have an object with many fields or an object graph and you only need to
-      calculate some information or need a subset of fields, you can use the `apply()`
-      method. With the `apply()` method, the whole object does not need to be sent over
-      the line; only the information that is relevant is sent.
+    The following are some considerations you need to know when you use
+    AtomicReference:
```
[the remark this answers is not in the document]
mdumandag: I feel like the current version is better

mdumandag:
@yuce I have gone over your comments and fixed most of them. I put comments on the ones that I didn't apply. 

BTW, can you do your further reviews in #518, as it is the base PR that we need to merge? I have applied your comments to there and just rebased here

yuce (approved):

utku-caglayan (approved):
LGTM!

mdumandag:
Thanks for the reviews guys!

PR #523, opened 2022-01-12
Make the tests more resilient to timing problems

mdumandag:
We had some test failures on the backward compatibility runs, mainly
due to the fact that, the failing tests were not resilient enough
to timing problems.

In the `HeartbeatTest::test_heartbeat_stopped_and_restored`, it was
possible that the responses coming from the server for the previous
heartbeat requests to override what we wrote as the `last_read_time`.

In the `ListenerAddMemberTest::test_add_member`, it was possible that
the `map#put` would complete before the listener is registered to the
newly added connection.

### DIFF
--- a/tests/integration/backward_compatible/heartbeat_test.py
+++ b/tests/integration/backward_compatible/heartbeat_test.py
@@ -1,3 +1,6 @@
+import threading
+import time
+
 from hazelcast import HazelcastClient
 from hazelcast.core import Address
 from tests.base import HazelcastTestCase
@@ -5,6 +8,8 @@ from tests.util import open_connection_to_address, wait_for_partition_table
 
 
 class HeartbeatTest(HazelcastTestCase):
+    rc = None
+
     @classmethod
     def setUpClass(cls):
         cls.rc = cls.create_rc()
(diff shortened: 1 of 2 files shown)


tests/integration/backward_compatible/heartbeat_test.py
```diff
@@ -48,25 +53,57 @@ def collector(c, *_):
...
+                    return
+
+                for connection in client._connection_manager.active_connections.values():
+                    if connection.remote_address == address:
+                        connection.last_read_time -= timeout
+                        break
+
+                time.sleep((i + 1) * 0.1)
+
+        t = threading.Thread(target=run)
```
srknzl, line 107: Ok, the logic seems very similar to my fix to HeartbeatFromServerTest in the Node.js client, but why do we need a thread and threading.Event here? Looks like you want the thread execution to wait something, but the event flag is set immediately after thread is started. So, it actually does not wait for anything in other threads, does it?

mdumandag: It is a minor thing, but I wanted the loop to start right before the assertion. I could remove it if you want, and it probably will not affect anything

srknzl: simulate_heartbeat_loss was starting right before the assertion before this change as well. Wasn't it? I think if threading is not necessary, it can be removed to make code less complex.

srknzl: Sorry looks like threading is necessary here but event can be removed. Right? Then, nevermind. It is minor thing

srknzl (approved):
Thanks looks good

tests/integration/backward_compatible/heartbeat_test.py
mdumandag: Yes. BTW, I have removed it nevertheless to make it simpler

tests/integration/backward_compatible/heartbeat_test.py
```diff
@@ -48,25 +53,55 @@ def collector(c, *_):
...
+            assertion_succeeded,
+        )
 
         def assert_heartbeat_stopped_and_restored():
-            self.assertEqual(1, len(connection_added_collector.connections))
-            self.assertEqual(1, len(connection_removed_collector.connections))
-            stopped_connection = connection_added_collector.connections[0]
-            restored_connection = connection_removed_collector.connections[0]
+            self.assertTrue(len(connection_added_collector.connections) >= 1)
+            self.assertTrue(len(connection_removed_collector.connections) >= 1)
```
yuce, line 66: `GreaterOrEqual` may also be used instead of `assertTrue+len`.

tests/integration/backward_compatible/heartbeat_test.py
```diff
@@ -48,25 +53,55 @@ def collector(c, *_):
             connection_added_collector, connection_removed_collector
         )
 
-        self.simulate_heartbeat_lost(self.client, addr, 2)
+        assertion_succeeded = [False]
```
yuce, line 56: Instead of making `assertion_succeeded` a list, we can use `nonlocal assertion_succeeded` in `assert_heartbeat_stopped_and_restored` to mutate its value.

tests/integration/backward_compatible/listener_test.py
```diff
@@ -74,13 +77,35 @@ def test_add_member(self, _, is_smart):
...
         client = self.create_client(self.client_config)
         random_map = client.get_map(random_string()).blocking()
-        random_map.add_entry_listener(added_func=self.collector)
+        random_map.add_entry_listener(added_func=self.collector, updated_func=self.collector)
         m2 = self.cluster.start_member()
         wait_for_partition_table(client)
         key_m2 = generate_key_owned_by_instance(client, m2.uuid)
-        random_map.put(key_m2, "value")
+
+        assertion_succeeded = [False]
```
yuce, line 85: Using `nonlocal` would make a better solution here too.

yuce (approved):
Just a few minor comments.

mdumandag:
thanks for the reviews

PR #529, opened 2022-02-28
[API-965] Implement SQL JSON support

### LINKED ISSUE #508: [TRACKING ISSUE] [HZ-563] JSON type support

See https://github.com/hazelcast/hazelcast/pull/19303 for details.

---

Adds `JSON` as a supported type (available in most contexts, aside from JDBC since that's not part of the SQL spec) as well as few functions from SQL JSON TR. 
Features: 
- `JSON_QUERY(jsonArg VARCHAR|JSON, jsonPath VARCHAR ... <extended syntax> ): JSON` - returns a json-object/array by given JSONPath (implemented according to SQL JSON TR)
- JSON_VALUE(jsonArg VARCHAR|JSON, jsonPath VARCHAR ... <extended syntax>): VARCHAR|<RETURNING TYPE>` - returns a primitive value by given JSONPath (implemented according to SQL JSON TR)
- `JSON_PARSE(jsonString VARCHAR): JSON` - workaround function to mitigate issues with aggressive `CAST(literal AS <type>)` optimizations, analyzer replaces any instance of `CAST(<varcharLiteral> AS JSON)` with a call of this function. 
- `CAST(x AS JSON)` support for `VARCHAR`, columns/literals and dynamic params are supported. 
- `JSON` as a first-class type, with IMap and Kafka Connector support, usable in most contexts, similarly to any other type.

mdumandag:
This PR adds implementation, tests, code samples, and documentation
for the SQL JSON column type.

Closes #508

### DIFF
--- a/docs/using_python_client_with_hazelcast.rst
+++ b/docs/using_python_client_with_hazelcast.rst
@@ -1598,7 +1598,7 @@ objects and queries it using nested fields, which is not possible with the
 
     client.sql.execute(
         """
-    CREATE MAPPING IF NOT EXISTS json_employees
+    CREATE OR REPLACE MAPPING json_employees
     TYPE IMap
     OPTIONS (
         'keyFormat' = 'int',
@@ -1679,7 +1679,7 @@ top-level fields:
 
     client.sql.execute(
         """
-    CREATE MAPPING IF NOT EXISTS json_flat_employees (
+    CREATE OR REPLACE MAPPING json_flat_employees (
         __key INT,
         name VARCHAR,
         salary INT
--- a/docs/using_python_client_with_hazelcast.rst
+++ b/docs/using_python_client_with_hazelcast.rst
@@ -1572,22 +1572,114 @@ Query parameters have the following benefits:
 Querying JSON Objects
 ~~~~~~~~~~~~~~~~~~~~~
 
-To query JSON objects, you should create an explicit mapping using the
+In Hazelcast, the SQL service supports the following ways of working with
+JSON data:
+
+- ``json``: Maps JSON data to a single column of ``JSON`` type where you can
+  use `JsonPath
+  <https://docs.hazelcast.com/hazelcast/latest/sql/working-with-json#querying-json>`__
+  syntax to query and filter it, including nested levels.
+- ``json-flat``: Maps JSON top-level fields to columns with non-JSON types
+  where you can query only top-level keys.
+
+**json**
+
+To query ``json`` objects, you should create an explicit mapping using the
 `CREATE MAPPING
 <https://docs.hazelcast.com/hazelcast/latest/sql/create-mapping.html>`__
 statement, similar to the example above.
 
 For example, this code snippet creates a mapping to a new map called
-``json_employees``, which stores the JSON values ``name`` and ``salary`` and
-query it:
+``json_employees``, which stores the JSON values as ``HazelcastJsonValue``
+objects and queries it using nested fields, which is not possible with the
+``json-flat`` type:
 
 .. code:: python
 
-    client = hazelcast.HazelcastClient()
+    client.sql.execute(
+        """
+    CREATE MAPPING IF NOT EXISTS json_employees
+    TYPE IMap
+    OPTIONS (
+        'keyFormat' = 'int',
+        'valueFormat' = 'json'
+    )
+        """
+    ).result()
+
+    json_employees = client.get_map("json_employees").blocking()
+
+    json_employees.set(
+        1,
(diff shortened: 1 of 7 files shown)


docs/using_python_client_with_hazelcast.rst
```diff
@@ -1572,22 +1572,114 @@ Query parameters have the following benefits:
...
+``json_employees``, which stores the JSON values as ``HazelcastJsonValue``
+objects and queries it using nested fields, which is not possible with the
+``json-flat`` type:
 
 .. code:: python
 
-    client = hazelcast.HazelcastClient()
+    client.sql.execute(
+        """
+    CREATE MAPPING IF NOT EXISTS json_employees
```
yuce, line 1601: CREATE OR REPLACE MAPPING would be better in case the user has a mapping created using an older example.

docs/using_python_client_with_hazelcast.rst
```diff
@@ -1572,22 +1572,114 @@ Query parameters have the following benefits:
...
+
+**json-flat**
+
+To query ``json-flat`` objects, you should create an explicit mapping using the
+`CREATE MAPPING
+<https://docs.hazelcast.com/hazelcast/latest/sql/create-mapping.html>`__
+statement, similar to the example above.
+
+For example, this code snippet creates a mapping to a new map called
+``json_flat_employees``, which stores the JSON values with columns ``name``
```
yuce, line 1674: Do you think using just `employees` would make the example simpler?

examples/sql/sql_json_example.py
```diff
@@ -0,0 +1,34 @@
+import hazelcast
```
yuce, line 1: How about enabling logging for the example? It increases the LOC but in case there is a problem, that would help the user to spot the problem.

examples/sql/sql_json_example.py
```diff
@@ -0,0 +1,34 @@
...
+    'valueFormat' = 'json'
+)
+    """
+).result()
+
+# Select the names of employees older than 25
+result = client.sql.execute(
+    """
+SELECT JSON_VALUE(this, '$.name') AS name
+FROM employees
```
yuce, line 29: Using just `select name from ...` should also work. Are you using `JSON_VALUE` to demonstrate its usage?

yuce (commented):
Looks good, just a few ideas.

docs/using_python_client_with_hazelcast.rst
mdumandag: I wanted to distinguish the map/mapping names if the users are going to test the two code snippets. So, `json_employees` and `json_flat_employees` are coming from there, and I don't think it makes the example harder, as it is just a name

examples/sql/sql_json_example.py
mdumandag: I just checked the other code samples, and apparently, we haven't enabled logging in any of them. But, I agree with you and I believe it would definitely be useful. So, if it is okay for you, I will create an issue for it, and will enable logging for all code samples in a later PR

examples/sql/sql_json_example.py
mdumandag: Yes

examples/sql/sql_json_example.py
yuce: Sure, thanks!

yuce (approved):
Looks good!

srknzl (changes requested):
Can you update the SQL limitations section? The following points are not valid anymore(they are not limitations anymore):

- No support for the JSON type. You can’t use functions such as JSON_VALUE or JSON_QUERY.

- No support for the CREATE INDEX statement. To create indexes for maps in Hazelcast, see the :func:`add_index() <hazelcast.proxy.map.Map.add_index>` method.

- No support for window functions. You cannot group or aggregate results in streaming queries.(See https://docs.hazelcast.com/hazelcast/5.1/sql/querying-streams#aggregation)

examples/sql/sql_json_example.py
srknzl: `select name from`

srknzl (approved):

mdumandag:
@yüce can I get a re-review from you?

examples/sql/sql_json_example.py
yuce: It seems the table has to be created with `json-flat` for `select name ...` to work.

yuce (approved):

examples/sql/sql_json_example.py
srknzl: yeap

PR #530, opened 2022-03-01
[API-1229] Add a configuration option to enable hostname verification

mdumandag:
This PR adds a new configuration element called `ssl_check_hostname`,
which has a similar name to the standard library [counterpart](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.check_hostname).

To maintain backward compatibility, the option is disabled by
default.

The hostname to use during the verification process is extracted
from the hostname of the configured member address.

This PR also adds an extensive test suite for the feature,
and documentation.


docs/securing_client_connection.rst
```diff
@@ -225,6 +226,28 @@ The following is an example configuration:
         ssl_ciphers="DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA"
     )
 
+Checking Hostname
+^^^^^^^^^^^^^^^^^
+
+During the TLS/SSL handshake, the client can verify that the hostname
+or the IP address of the member matches with the information provided
+in the member's certificate's Subject Alternative Name extension or
```
yuce, line 234: How about "... in the Subject Alternative Name extension of the member's certificate ..."?

docs/securing_client_connection.rst
```diff
@@ -225,6 +226,28 @@ The following is an example configuration:
...
 
+Checking Hostname
+^^^^^^^^^^^^^^^^^
+
+During the TLS/SSL handshake, the client can verify that the hostname
+or the IP address of the member matches with the information provided
+in the member's certificate's Subject Alternative Name extension or 
+Common Name field in the Subject field.
+
+The hostname the client uses during the verification process is the
```
yuce, line 237: How about "... Hostname used during ..."

docs/securing_client_connection.rst
```diff
@@ -225,6 +226,28 @@ The following is an example configuration:
...
+hostname of the configured member address in the client constructor.
+
+By default, hostname verification is disabled, but it is highly
```
yuce, line 240: I think you wanted to mention Python 3.7+ requirement in the doc.

hazelcast/reactor.py
```diff
@@ -533,7 +529,7 @@ def _set_socket_options(self, config):
 
             self.socket.setsockopt(level, option_name, value)
 
-    def _wrap_as_ssl_socket(self, config):
+    def _wrap_as_ssl_socket(self, config, hostname: str):
```
yuce, line 532: Would adding the type of `config` be useful here?

tests/integration/backward_compatible/ssl_tests/hostname_verification/ssl_hostname_verification_test.py
```diff
@@ -0,0 +1,128 @@
...
+            self.start_client_with("tls-host-no-entry.pem", "127.0.0.1:5701")
+
+    def test_hostname_verification_disabled(self):
+        # When hostname verification is disabled, the scenarious that
+        # would fail in `test_hostname_verification_with_no_entry` will
+        # no longer fail, showing that it is working as expected.
+        self.start_member_with("tls-host-no-entry.p12")
+
+        self.start_client_with("tls-host-no-entry.pem", "localhost:5701", check_hostname=False)
+        self.start_client_with("tls-host-no-entry.pem", "127.0.0.1:5701", check_hostname=False)
```
yuce, line 104: Do you think we can use table driven tests to reduce the test code? https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest

mdumandag: I took a look at it and it looks cool but, I decided not to change the tests to that because I think it decreases the readability and our ability to debug the tests. Besides, I have some comments on the tests that explain the test a little bit which could be helpful for future readers.

yuce (approved):
Looks good!

tests/integration/backward_compatible/ssl_tests/hostname_verification/ssl_hostname_verification_test.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def test_hostname_verification_with_different_san(self):
+        # There is a valid entry, but it does not match with the address of the member.
+        self.start_member_with("tls-host-not-our-san.p12")
+
+        with self.assertRaisesRegex(IllegalStateError, "Unable to connect to any cluster"):
+            self.start_client_with("tls-host-not-our-san.pem", "localhost:5701")
+
+        with self.assertRaisesRegex(IllegalStateError, "Unable to connect to any cluster"):
+            self.start_client_with("tls-host-loopback-san-dns.pem", "127.0.0.1:5701")
```
srknzl, line 79: Did you do a copy paste mistake here? 
```suggestion
            self.start_client_with("tls-host-not-our-san.pem", "127.0.0.1:5701")
```

srknzl, line 79: I feel like trying all these scenarios might be unnecessary and might be testing python core functionality. If you agree feel free to leave single test with a simpler setup.

mdumandag: Yes, thanks for catching this. I have updated the tests to re-use the file name to avoid such mistakes now.

mdumandag: I copied the tests from the Java code, and I believe they are important to make sure that we have the same behavior with the Java client. These are all short-running tests, so I don't see a problem in having them. But, yes you are right, we are essentially testing the behavior of openssl

srknzl (approved):
LGTM

tests/integration/backward_compatible/ssl_tests/hostname_verification/ssl_hostname_verification_test.py
srknzl: Ok then

yuce (approved):

PR #531, opened 2022-03-08
[API-521] Add support for compact serialization to public APIs

mdumandag:
For compact serialization to work, we have to control the results
of the serialization. Also, for the responses we got from the server,
we have to eagerly deserialize them, and retry after receiving the
schema from the cluster.

With this PR, we make sure that on each API we support compact serialization,
we check the result of the `to_data` call to see if it throws or not.

If it throws, we try to send the schema to the cluster and retry.

Also, this PR adds tests for all the APIs we support compact serialization.
For now, we didn't implement eager deserialization for the APIs that
return `ImmutableLazyDataList`, so there are no tests for such APIs.

To be able to add those tests, we would need quite a bit of compact
serializable classes written in Java. This PR includes aforementioned
Java code, but they might be moved to the remote controller package so
that it would be accessible to all clients.

### DIFF
--- a/hazelcast/listener.py
+++ b/hazelcast/listener.py
@@ -147,7 +147,7 @@ class ListenerService:
             try:
                 handler(message)
             except SchemaNotFoundError as e:
-                self._fetch_schema_and_handle_again(e, handler, message, correlation_id)
+                self._fetch_schema_and_handle_again(e, handler, message)
         else:
             _logger.debug("Got event message with unknown correlation id: %s", message)
 
@@ -156,7 +156,6 @@ class ListenerService:
         error: SchemaNotFoundError,
         handler: typing.Callable[[InboundMessage], None],
         message: InboundMessage,
-        correlation_id: int,
     ) -> None:
         schema_id = error.schema_id
 
@@ -171,7 +170,10 @@ class ListenerService:
                 return
 
             message.reset_next_frame()
-            self.handle_client_message(message, correlation_id)
+            try:
+                handler(message)
+            except SchemaNotFoundError as e:
+                self._fetch_schema_and_handle_again(e, handler, message)
 
         fetch_schema_future = self._compact_schema_service.fetch_schema(schema_id)
         fetch_schema_future.add_done_callback(callback)
--- a/hazelcast/partition.py
+++ b/hazelcast/partition.py
@@ -60,7 +60,7 @@ class PartitionService:
         try:
             key_data = self._serialization_service.to_data(key)
         except SchemaNotReplicatedError as e:
-            self._send_schema_and_retry_fn(e, lambda: ...).result()
+            self._send_schema_and_retry_fn(e, lambda: None).result()
             return self.get_partition_id(key)
 
         return self._service.get_partition_id(key_data)
--- a/hazelcast/partition.py
+++ b/hazelcast/partition.py
@@ -5,6 +5,7 @@ import typing
 
 from hazelcast.errors import ClientOfflineError
 from hazelcast.hash import hash_to_index
+from hazelcast.serialization.compact import SchemaNotReplicatedError
 
 _logger = logging.getLogger(__name__)
 
@@ -27,11 +28,12 @@ class PartitionService:
     owner or the partition id of a key.
     """
 
-    __slots__ = ("_service", "_serialization_service")
+    __slots__ = ("_service", "_serialization_service", "_send_schema_and_retry_fn")
 
-    def __init__(self, internal_partition_service, serialization_service):
+    def __init__(self, internal_partition_service, serialization_service, send_schema_and_retry_fn):
         self._service = internal_partition_service
         self._serialization_service = serialization_service
+        self._send_schema_and_retry_fn = send_schema_and_retry_fn
 
     def get_partition_owner(self, partition_id: int) -> typing.Optional[uuid.UUID]:
         """
@@ -55,7 +57,12 @@ class PartitionService:
         Returns:
             The partition id.
         """
-        key_data = self._serialization_service.to_data(key)
+        try:
+            key_data = self._serialization_service.to_data(key)
+        except SchemaNotReplicatedError as e:
+            self._send_schema_and_retry_fn(e, lambda: ...).result()
+            return self.get_partition_id(key)
+
         return self._service.get_partition_id(key_data)
 
     def get_partition_count(self) -> int:
--- a/hazelcast/proxy/base.py
+++ b/hazelcast/proxy/base.py
@@ -256,35 +256,22 @@ class EntryEvent(typing.Generic[KeyType, ValueType]):
 
 
 class TopicMessage(typing.Generic[MessageType]):
-    """Topic message."""
+    """Topic message.
 
-    __slots__ = ("_name", "_message", "_publish_time", "_member")
+    Attributes:
+        name: Name of the proxy that fired the event.
+        message: The message sent to Topic.
+        publish_time: UNIX time that the event is published as seconds.
+        member: Member that fired the event.
+    """
 
-    def __init__(self, name, message, publish_time, member):
+    __slots__ = ("name", "message", "publish_time", "member")
+
+    def __init__(self, name: str, message: MessageType, publish_time: int, member: MemberInfo):
         self._name = name
         self._message = message
         self._publish_time = publish_time
         self._member = member
-
-    @property
-    def name(self) -> str:
-        """Name of the proxy that fired the event."""
-        return self._name
-
-    @property
-    def publish_time(self) -> int:
-        """UNIX time that the event is published as seconds."""
-        return self._publish_time
-
-    @property
-    def member(self) -> MemberInfo:
-        """Member that fired the event."""
-        return self._member
-
-    @property
-    def message(self) -> MessageType:
-        """The message sent to Topic."""
-        return self._message
 
     def __repr__(self):
         return "TopicMessage(message=%s, publish_time=%s, topic_name=%s, publishing_member=%s)" % (
--- a/hazelcast/proxy/base.py
+++ b/hazelcast/proxy/base.py
@@ -107,10 +107,9 @@ class TransactionalProxy:
         self._to_object = serialization_service.to_object
         self._to_data = serialization_service.to_data
         self._send_schema_and_retry = context.compact_schema_service.send_schema_and_retry
-        self._no_op_func = lambda: None
 
     def _send_schema(self, error):
-        return self._send_schema_and_retry(error, self._no_op_func).result()
+        return self._send_schema_and_retry(error, lambda: None).result()
 
     def _invoke(self, request, response_handler=_no_op_response_handler):
         invocation = Invocation(
--- a/hazelcast/proxy/base.py
+++ b/hazelcast/proxy/base.py
@@ -268,10 +268,10 @@ class TopicMessage(typing.Generic[MessageType]):
     __slots__ = ("name", "message", "publish_time", "member")
 
     def __init__(self, name: str, message: MessageType, publish_time: int, member: MemberInfo):
-        self._name = name
-        self._message = message
-        self._publish_time = publish_time
-        self._member = member
+        self.name = name
+        self.message = message
+        self.publish_time = publish_time
+        self.member = member
 
     def __repr__(self):
         return "TopicMessage(message=%s, publish_time=%s, topic_name=%s, publishing_member=%s)" % (
--- a/hazelcast/proxy/base.py
+++ b/hazelcast/proxy/base.py
@@ -3,11 +3,9 @@ import typing
 import uuid
 
 from hazelcast.core import MemberInfo
-from hazelcast.future import Future
+from hazelcast.types import KeyType, ValueType, ItemType, MessageType, BlockingProxyType
 from hazelcast.invocation import Invocation
 from hazelcast.partition import string_partition_strategy
-from hazelcast.serialization.compact import SchemaNotReplicatedError
-from hazelcast.types import KeyType, ValueType, ItemType, MessageType, BlockingProxyType
 from hazelcast.util import get_attr_name
 
 MAX_SIZE = float("inf")
@@ -33,20 +31,7 @@ class Proxy(typing.Generic[BlockingProxyType], abc.ABC):
         self._register_listener = listener_service.register_listener
         self._deregister_listener = listener_service.deregister_listener
         self._is_smart = context.config.smart_routing
-        self._send_compact_schema = context.compact_schema_service.send_schema
-
-    def _send_schema_and_retry(
-        self,
-        error: SchemaNotReplicatedError,
-        func: typing.Callable[..., Future],
-        *args: typing.Any,
-        **kwargs: typing.Any
-    ) -> Future:
-        def continuation(future):
-            future.result()
-            return func(*args, **kwargs)
-
-        return self._send_compact_schema(error.schema, error.clazz).continue_with(continuation)
+        self._send_schema_and_retry = context.compact_schema_service.send_schema_and_retry
 
     def destroy(self) -> bool:
         """Destroys this proxy.
@@ -121,6 +106,11 @@ class TransactionalProxy:
         serialization_service = context.serialization_service
         self._to_object = serialization_service.to_object
         self._to_data = serialization_service.to_data
+        self._send_schema_and_retry = context.compact_schema_service.send_schema_and_retry
+        self._no_op_func = lambda: None
+
+    def _send_schema(self, error):
+        return self._send_schema_and_retry(error, self._no_op_func).result()
 
     def _invoke(self, request, response_handler=_no_op_response_handler):
         invocation = Invocation(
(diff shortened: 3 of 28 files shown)


hazelcast/serialization/compact.py
```diff
@@ -0,0 +1,2240 @@
...
+    def read(self, inp: ObjectDataInput) -> typing.Any:
+        schema_id = inp.read_long()
+        schema = self._id_to_schema.get(schema_id)
+        if not schema:
+            raise SchemaNotFoundError(schema_id)
+
+        serializer = self._type_name_to_serializer.get(schema.type_name)
+        if not serializer:
+            raise HazelcastSerializationError(
+                f"No serializer is registered for the type name '{schema.type_name}'"
```
srknzl, line 72: Maybe saying `No compact serializer is registered` is better for users to understand the data is in compact format.

srknzl:
Why CI did not run? Also, can you resolve conflicts ?

hazelcast/proxy/map.py
```diff
@@ -1421,6 +1611,12 @@ def values(self, predicate: Predicate = None) -> Future[typing.List[ValueType]]:
         """
         if predicate:
             if isinstance(predicate, PagingPredicate):
+                try:
+                    predicate.iteration_type = IterationType.VALUE
```
srknzl, line 1615: We can get this line out of try block

hazelcast/partition.py
```diff
@@ -55,7 +57,12 @@ def get_partition_id(self, key: typing.Any) -> int:
         Returns:
             The partition id.
         """
-        key_data = self._serialization_service.to_data(key)
+        try:
+            key_data = self._serialization_service.to_data(key)
+        except SchemaNotReplicatedError as e:
+            self._send_schema_and_retry_fn(e, lambda: ...).result()
+            return self.get_partition_id(key)
```
srknzl, line 64: What do you think about doing what I did in Node.js? I mean we don't need to send the schema to the cluster for this operation because the data is not sent to the cluster. It will affect the performance of this method but don't know how much important it is.

hazelcast/sql.py
```diff
@@ -1233,15 +1241,18 @@ def execute(self, sql, params, cursor_buffer_size, timeout, expected_result_type
 
         connection = None
         try:
+            try:
+                # Serialize the passed parameters.
+                serialized_params = [
+                    self._serialization_service.to_data(param) for param in statement.parameters
+                ]
+            except SchemaNotReplicatedError as e:
+                return self._send_schema_and_retry_fn(e, self.execute, sql, params, kwargs)
```
srknzl, line 1250: Is this `kwargs` usage valid? 

```python3
def a(a, b, c):
     print(kwargs)

a(a = 1, b = 2, c = 3)
```
throws
```
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in a
NameError: name 'kwargs' is not defined

```

To verify, can you add SQL compact tests?

hazelcast/proxy/queue.py
```diff
@@ -176,7 +187,8 @@ def drain_to(self, target_list: typing.List[ItemType], max_size: int = -1) -> Fu
 
         def handler(message):
             response = queue_drain_to_max_size_codec.decode_response(message)
-            target_list.extend(map(self._to_object, response))
+            items = [self._to_object(item) for item in response]
+            target_list.extend(items)
```
srknzl, line 191: I suppose there may be extra list allocation here. map() returns an iterable and extend will use it to extend the existing list without creating a new one. You create another one here. Why do you do this change though?

srknzl (commented):
This is the half of my review

hazelcast/proxy/base.py
```diff
@@ -105,6 +106,11 @@ def __init__(self, name, transaction, context):
         serialization_service = context.serialization_service
         self._to_object = serialization_service.to_object
         self._to_data = serialization_service.to_data
+        self._send_schema_and_retry = context.compact_schema_service.send_schema_and_retry
+        self._no_op_func = lambda: None
```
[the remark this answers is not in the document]
mdumandag: No, the transactional proxies only have blocking APIs, so calling a blocking API on the callback of the send schema operation would deadlock the reactor thread

hazelcast/proxy/cp/atomic_long.py
```diff
@@ -157,7 +158,11 @@ def alter(self, function: typing.Any) -> Future[None]:
             function: The function that alters the currently stored value.
         """
         check_not_none(function, "Function cannot be None")
-        function_data = self._to_data(function)
+        try:
+            function_data = self._to_data(function)
+        except SchemaNotReplicatedError as e:
+            return self._send_schema_and_retry(e, self.alter, function)
```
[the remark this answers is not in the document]
mdumandag: I think this one is OK and shows the purpose better

hazelcast/proxy/map.py
```diff
@@ -210,14 +265,19 @@ def add_entry_listener(
...
             event = EntryEvent(
-                self._to_object,
-                key_,
-                value,
-                old_value,
-                merging_value,
+                self._to_object(key_data),
+                self._to_object(value_data),
+                self._to_object(old_value_data),
+                self._to_object(merging_value_data),
```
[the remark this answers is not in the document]
mdumandag: In the listener service, we have a support for it. I will send a fix for the problem we discussed (about removing handlers while fetching the schema), and you can take a look around it to see the code

hazelcast/partition.py
mdumandag: I think this method is not important, and I don't want to introduce a new parameter to to_data

hazelcast/sql.py
mdumandag: no, not at all. it was most probably a copy-paste error

hazelcast/proxy/queue.py
mdumandag: I need to verify that for all the items in the response, there is a schema in local. So, I have to call to_object on each of them before starting to extend the list, to avoid duplicate items in case of SchemaNotFound errors

hazelcast/sql.py
mdumandag: I will improve the SQL compatibility tests, but I need to update the class in the remote controller. I will do that today

hazelcast/listener.py
```diff
@@ -171,7 +170,10 @@ def callback(future):
                 return
 
             message.reset_next_frame()
-            self.handle_client_message(message, correlation_id)
+            try:
+                handler(message)
+            except SchemaNotFoundError as e:
+                self._fetch_schema_and_handle_again(e, handler, message)
```
srknzl, line 176: Do we need to recursively call this function? I think it is enough to try fetching the schema once. Let's say somehow schema is removed from the cluster, this event handler will be tried to be called forever.

hazelcast/proxy/base.py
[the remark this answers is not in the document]
mdumandag: No, it does not have to return a Future. In our implementation, continue_with behaves differently according to the return value of the callback. If it is returning a Future, it chains it and resolves when that Future resolves, if it returns a normal value, it resolves immediately when the Future we are continue_with resolves. See https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/future.py#L152-L155

hazelcast/listener.py
mdumandag: Schema cannot be removed from the cluster. Besides, you have to try recursively to deal with scenarios like
```
def event_handler(...)
  event_message = ...
  key = to_object(event_message["key_data"])
  value = to_object(event_message["value_data"])
```
where key and value are separate compact objects, and we don't have a schema for them

hazelcast/sql.py
mdumandag: OK, I have updated the test, and found a way to use serialization on the client side without requiring a change on the remote controller classes.

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1691 @@
...
+
+    def test_put(self):
+        with self.transaction:
+            transactional_multi_map = self._get_transactional_multi_map()
+            self.assertTrue(
+                transactional_multi_map.put(INNER_COMPACT_INSTANCE, OUTER_COMPACT_INSTANCE)
+            )
+
+    def test_get(self):
+        self.multi_map.put(OUTER_COMPACT_INSTANCE, INNER_COMPACT_INSTANCE)
```
srknzl, line 1533: We can add a comment about why we put from the same client.

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1691 @@
...
+        map_name = random_string()
+        self.client_config = copy.deepcopy(NearCachedMapCompactCompatibilityTest.client_config)
+        self.client_config["near_caches"] = {map_name: {}}
+        super().setUp()
+        self.map = self.client.get_map(map_name).blocking()
+
+    def test_get_for_near_cache(self):
+        # Another variant of the test in the superclass, where we lookup a key
+        # which has a value whose schema is not fully sent to the
+        # cluster. The near cache will try to serialize it, but it should
```
srknzl, line 839: > which has a value whose schema is not fully sent to the cluster.

The value here is the `OUTER_COMPACT_INSTANCE`. What do you mean by schema not fully sent to the cluster? Does not  `_put_from_another_client` send its schema to the cluster ?

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1691 @@
...
+        self.reliable_topic.destroy()
+        super().tearDown()
+
+    def test_publish(self):
+        messages = []
+
+        def listener(message):
+            messages.append(message)
+
+        self.reliable_topic.add_listener(listener)
```
srknzl, line 1084: In order to test deserialization of compact object in the TopicMessage, can we add the listener from another client?

hazelcast/listener.py
mdumandag: Maybe later, the current logic both in the client and server is around fetching single schemas. We can do something like this maybe in 5.2

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1684 @@
...
+class MapCompatibilityTest(CompactCompatibilityBase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.map = self.client.get_map(random_string()).blocking()
+
+    def tearDown(self) -> None:
+        self.map.destroy()
+        super().tearDown()
+
+    def test_add_entry_listener_with_key_and_predicate(self):
```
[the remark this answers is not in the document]
mdumandag: Yeah, maybe we could have disallow such usage but I believe we need to do that first in Java

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1684 @@
...
+    def test_offer(self):
+        self.assertTrue(self.queue.offer(OUTER_COMPACT_INSTANCE))
+
+    def test_peek(self):
+        self._add_from_another_client(OUTER_COMPACT_INSTANCE)
+        self.assertEqual(OUTER_COMPACT_INSTANCE, self.queue.peek())
+
+    def test_poll(self):
+        self._add_from_another_client(OUTER_COMPACT_INSTANCE)
+        self.assertEqual(OUTER_COMPACT_INSTANCE, self.queue.poll())
```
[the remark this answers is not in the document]
mdumandag: This test is not just for to_data, I have written tests for places I call to_object as well

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1684 @@
...
+    def test_add_all(self):
+        self.assertEqual(
+            1, self.ringbuffer.add_all([INNER_COMPACT_INSTANCE, OUTER_COMPACT_INSTANCE])
+        )
+        self.assertEqual(INNER_COMPACT_INSTANCE, self.ringbuffer.read_one(0))
+        self.assertEqual(OUTER_COMPACT_INSTANCE, self.ringbuffer.read_one(1))
+
+    def test_read_one(self):
+        self._add_from_another_client(OUTER_COMPACT_INSTANCE)
+        self.assertEqual(OUTER_COMPACT_INSTANCE, self.ringbuffer.read_one(0))
```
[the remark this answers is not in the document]
mdumandag: See the reasoning above

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1684 @@
...
+        self._add_from_another_client(OUTER_COMPACT_INSTANCE)
+        with self.transaction:
+            transactional_queue = self._get_transactional_queue()
+            self.assertEqual(OUTER_COMPACT_INSTANCE, transactional_queue.poll())
+
+    def test_peek(self):
+        self._add_from_another_client(OUTER_COMPACT_INSTANCE)
+        with self.transaction:
+            transactional_queue = self._get_transactional_queue()
+            self.assertEqual(OUTER_COMPACT_INSTANCE, transactional_queue.peek())
```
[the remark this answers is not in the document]
mdumandag: See the reasoning above

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1684 @@
...
+
+    def test_apply(self):
+        self.assertEqual(
+            OUTER_COMPACT_INSTANCE,
+            self.atomic_reference.apply(CompactReturningFunction()),
+        )
+
+    def test_get(self):
+        self.atomic_reference.set(OUTER_COMPACT_INSTANCE)
+        self.assertEqual(OUTER_COMPACT_INSTANCE, self.atomic_reference.get())
```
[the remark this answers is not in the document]
mdumandag: See the reasoning above

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
```diff
@@ -0,0 +1,1691 @@
...
+        self.map.put(OUTER_COMPACT_INSTANCE, INNER_COMPACT_INSTANCE)
+        self.assertEqual(
+            [OUTER_COMPACT_INSTANCE],
+            self.map.key_set(paging(CompactPredicate(), 1)),
+        )
+
+    def test_load_all_with_keys(self):
+        try:
+            self.map.load_all(keys=[INNER_COMPACT_INSTANCE, OUTER_COMPACT_INSTANCE])
+        except NullPointerError:
```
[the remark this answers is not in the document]
mdumandag: Not really a big deal but I believe this should be OK

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
mdumandag: I meant the schema of the outer compact is not sent to the cluster from this client. let me add this to the comment as well

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
mdumandag: ReliableTopic uses lazy deserialization internally. It reads messages from a ringbuffer in batches. So, we currently don't fully support what you mentioned, hence there is no test for it

hazelcast/proxy/cp/atomic_long.py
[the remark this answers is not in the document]
mdumandag: I disagree. I think this repetition is OK. It reads better, it is consistent with every other API, so if there is no strong urge from you guys, I would like to keep this and others as they are

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
srknzl: Can't we support compact objects? I have seen the Node.js code and we can catch SchemaNotFoundError there and fetch the schema. Because there is a promise there.  https://github.com/hazelcast/hazelcast-nodejs-client/blob/061c096a88568d8b99e798259376ce673e5a8b03/src/proxy/topic/ReliableTopicListenerRunner.ts#L65

It may be similar in python

hazelcast/proxy/cp/atomic_long.py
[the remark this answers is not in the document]
srknzl: You can rename the method so that it reads good as well but it is not a must. I am fine with both.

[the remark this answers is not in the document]
mdumandag: Also, I don't think it is possible to implement it nicely. Let's say we extracted the logic to a method. What we will do? See the return path on the except block. How can we represent that in a separate method? We would need a try except again

tests/integration/backward_compatible/serialization/compact_compatibility/compact_compatibility_test.py
mdumandag: I went over the code and I believe although it is possible to do that, I believe it complicates the code a bit and I feel like we are solving the problem at the wrong place. It should have been fixed with the effort around lazy-deserialization.

So, I would like to skip it. WDYT @yuce @srknzl

srknzl: I also vote for skipping this for now. But we should document what won't work with compact objects with unknown schemas (places we do lazy deserialization)

srknzl (approved):

yuce (approved):

PR #534, opened 2022-03-11
Remove duplicate type hints

mdumandag:
Formerly, we had type hints in the docstrings, but after
introducing the actual type hints, they are not needed anymore.

This PR removes those duplicate type hints in the docstrings from
the public APIs.


yuce (dismissed):
Looks good!

utku-caglayan (approved):
Looks good!

PR #535, opened 2022-03-14
Fix mypy errors

mdumandag:
There were around 60 errors shown in the mypy checks about the various
minor problems across the client code base. This PR corrects
those errors and adds a check for the mypy to PR builder workflow.

Note that, for now, we are only checking the type hints in the
hazelcast directory (client's code base). Tests and examples are not
passing the mypy checks for now. We can improve this incrementally,
and add them in the future.

Also, we are not using the strict flag of the mypy, due to not
type hinting the internals yet. Hopefully, we will have that
in the future.

### DIFF
--- a/hazelcast/connection.py
+++ b/hazelcast/connection.py
@@ -280,7 +280,7 @@ class ConnectionManager:
 
         # Need to create copy of connection values to avoid modification errors on runtime
         for connection in list(self.active_connections.values()):
-            connection.close("Hazelcast client is shutting down", None)
+            connection.close_connection("Hazelcast client is shutting down", None)
 
         self.active_connections.clear()
         del self._connection_listeners[:]
@@ -613,9 +613,9 @@ class ConnectionManager:
     def _on_auth(self, response, connection):
         try:
             response = client_authentication_codec.decode_response(response.result())
-        except Exception as err:
-            connection.close("Failed to authenticate connection", err)
-            raise err
+        except Exception as e:
+            connection.close_connection("Failed to authenticate connection", e)
+            raise e
 
         status = response["status"]
         if status == _AuthenticationStatus.AUTHENTICATED:
@@ -632,7 +632,7 @@ class ConnectionManager:
                 "Authentication status code not supported. status: %s" % status
             )
 
-        connection.close("Failed to authenticate connection", err)
+        connection.close_connection("Failed to authenticate connection", err)
         raise err
 
     def _handle_successful_auth(self, response, connection):
@@ -649,7 +649,7 @@ class ConnectionManager:
 
             existing = self.active_connections.get(remote_uuid, None)
             if existing:
-                connection.close(
+                connection.close_connection(
                     "Duplicate connection to same member with UUID: %s" % remote_uuid, None
                 )
                 return existing
@@ -750,7 +750,7 @@ class ConnectionManager:
             # we can operate on. In those scenarios, we rely on the fact that we will
             # reopen the connections.
             reason = "Connection does not belong to the cluster %s" % self._cluster_id
-            connection.close(reason, None)
+            connection.close_connection(reason, None)
             raise ValueError(reason)
 
     def _on_cluster_restart(self):
@@ -824,7 +824,7 @@ class _HeartbeatManager:
 
         if (now - connection.last_read_time) > self._heartbeat_timeout:
             _logger.warning("Heartbeat failed over the connection: %s", connection)
-            connection.close(
+            connection.close_connection(
                 "Heartbeat timed out",
                 TargetDisconnectedError("Heartbeat timed out to connection %s" % connection),
             )
@@ -950,7 +950,8 @@ class Connection:
         self._write(message.buf)
         return True
 
-    def close(self, reason, cause):
+    # Not named close to distinguish it from the asyncore.dispatcher.close.
+    def close_connection(self, reason, cause):
         """Closes the connection.
 
         Args:
--- a/hazelcast/future.py
+++ b/hazelcast/future.py
@@ -61,7 +61,10 @@ class Future(typing.Generic[ResultType]):
         self._event.wait()
         if self._exception:
             re_raise(self._exception, self._traceback)
-        return self._result
+
+        # Result will be set to the correct type before we
+        # return from here
+        return self._result  # type: ignore[return-value]
 
     def _reactor_check(self):
         if not self.done() and hasattr(self._threading_locals, "is_reactor_thread"):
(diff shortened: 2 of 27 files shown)


hazelcast/future.py
```diff
@@ -307,6 +310,9 @@ def done(future, index):
     for index, future in enumerate(futures):
         # Capture the index in the closure or else we
         # will only update the last element.
-        future.add_done_callback(lambda f, captured_index=index: done(f, captured_index))
+        def callback(f, captured_index=index):
+            done(f, captured_index)
+
+        future.add_done_callback(callback)
```
yuce, line 316: How about something like:
```
def make_callback(index):
    return lambda f: done(f, index)

for index, future in enumerate(futures):
        future.add_done_callback(make_callback(index))
```

hazelcast/proxy/reliable_topic.py
```diff
@@ -582,7 +582,7 @@ def publish(self, message: MessageType) -> Future[None]:
             return self._add_or_fail(topic_message)
         elif overload_policy == TopicOverloadPolicy.DISCARD_OLDEST:
             return self._add_or_overwrite(topic_message)
-        elif overload_policy == TopicOverloadPolicy.DISCARD_NEWEST:
+        else:
```
yuce, line 585: Would it make sense to restore `elif overload_policy == TopicOverloadPolicy.DISCARD_NEWEST` here and add an `else` which raises an exception in case an unexpected policy is set?

hazelcast/proxy/replicated_map.py
```diff
@@ -78,21 +78,33 @@ def add_entry_listener(
             A registration id which is used as a key to remove the listener.
         """
         if key and predicate:
-            codec = replicated_map_add_entry_listener_to_key_with_predicate_codec
             key_data = self._to_data(key)
             predicate_data = self._to_data(predicate)
-            request = codec.encode_request(self.name, key_data, predicate_data, self._is_smart)
+            with_key_and_predicate_codec = (
+                replicated_map_add_entry_listener_to_key_with_predicate_codec
+            )
```
yuce, line 85: Are the parentheses required for keeping the line below some length?

hazelcast/proxy/replicated_map.py
```diff
@@ -78,21 +78,33 @@ def add_entry_listener(
             A registration id which is used as a key to remove the listener.
         """
         if key and predicate:
```
yuce, line 80: The `key` comparisons in this part don't look correct to me. What if `key == 0`?

hazelcast/reactor.py
```diff
@@ -22,14 +23,11 @@
 
 try:
     import fcntl
-except ImportError:
-    fcntl = None
 
-try:
-    from _thread import get_ident
+    _FCNTL_EXISTS = True
```
yuce, line 27: It may be nice to add a note about the platforms which `fcntl` is expected to be available.

yuce (changes requested):
Thanks for the PR! Here are a few ideas...

hazelcast/connection.py
```diff
@@ -863,7 +863,8 @@ def send_message(self, message):
         self._write(message.buf)
         return True
 
-    def close(self, reason, cause):
+    # Not named close to distinguish it from the asyncore.dispatcher.close.
+    def close_connection(self, reason, cause):
```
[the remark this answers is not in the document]
yuce: I would prefer `close_connection` over `close0`. If `disconnect` is not acceptable, I think it's OK to keep `close_connection`.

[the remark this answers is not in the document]
mdumandag: Ok, changed it back to close_connection. I believe we should have the word close in the method name

yuce (approved):
Looks good!

yuce (approved):

hazelcast/proxy/cp/atomic_reference.py
```diff
@@ -54,7 +54,9 @@ class AtomicReference(BaseCPProxy["BlockingAtomicReference"], typing.Generic[Ele
     server-side setting.
     """
 
-    def compare_and_set(self, expect: ElementType, update: ElementType) -> Future[bool]:
+    def compare_and_set(
+        self, expect: typing.Optional[ElementType], update: typing.Optional[ElementType]
```
srknzl, line 58: Why did we change these? These are also public API

hazelcast/proxy/map.py
```diff
@@ -186,7 +187,7 @@ def add_entry_listener(
             LOADED=loaded_func,
         )
 
-        if key and predicate:
+        if key is not None and predicate:
```
srknzl, line 190: Why do we have this asymmetry? key and predicate either should be allowed to be null or not.

hazelcast/proxy/reliable_topic.py
```diff
@@ -587,6 +587,8 @@ def publish(self, message: MessageType) -> Future[None]:
             return self._add_or_overwrite(topic_message)
         elif overload_policy == TopicOverloadPolicy.DISCARD_NEWEST:
             return self._add_or_discard(topic_message)
+        else:
+            raise ValueError(f"Unexpected overload policy is passed {overload_policy}")
```
srknzl, line 591: Probably we are checking this in config validation but it is to make type checker happy

hazelcast/proxy/replicated_map.py
```diff
@@ -78,7 +78,7 @@ def add_entry_listener(
         Returns:
             A registration id which is used as a key to remove the listener.
         """
-        if key and predicate:
+        if key is not None and predicate:
```
srknzl, line 81: Similar comment about asymmetry

srknzl (approved):
approving since the comments are not too critical

hazelcast/proxy/cp/atomic_reference.py
mdumandag: The type hints are introduced in this release, so not a breaking change. These were all optional (as it is now) but I missed that initially

hazelcast/proxy/map.py
mdumandag: predicate can be null or be a valid predicate. it can't be for example 0 or False, so there is no need for is None checks for it

hazelcast/proxy/reliable_topic.py
mdumandag: Yes

hazelcast/proxy/map.py
srknzl: I think predicates are not allowed to be null in Java. So you should change the default None value to a sentinel.

PR #539, opened 2022-03-27
New coverage yaml [API-1273]

alisengul53:
This PR includes a `coverage_runner.yaml` to run all tests including enterprise tests for PRs.
It changes the strategy of trigger mechanism. Normally it was triggered by `pull_request` event type. `pull_request` event type is running workflow in the coming PR. Now in the new `coverage_runner.yaml` `pull_request_target` is used. This event type is running the workflow in the base branch of which PR is opened. So it can reach repository secrets and run all tests. There are some security concerns and they are handled by checking the membership of the PR owner for the hazelcast organization.

The old `test_runner.yaml` is deleted. That is why some expected job runs are not executed.

This workflow has 3 options to run;

- A PR coming from a user in hazelcast organization. It will be triggered automatically
- When a PR is merged. It will be triggered automatically
- When a PR comes from community(not a member of hazelcast organization) and a member of hazelcast organization triggered it manually

When a PR comes, there is a membership check. It checks if the user is a member of hazelcast organization. Depending on the result of this membership check, the test runner job is working. It uses [this](https://github.com/hazelcast/hazelcast-tpm/blob/main/membership/action.yaml) action for membeship check.


kwart (approved):
Even though the risks of using the `pull_request_target` trigger are higher than for the `pull_request` one, I think the checks (implemented in the PR) keep us on the relatively safe side.

I didn't check the logic of the workflow, I only looked at the PR from the security perspective.

alisengul53:
Thanks for your review @kwart

srknzl:
> The old test_runner.yaml is deleted. That is why some expected job runs are not executed.

This seems very odd to me. The new workflow should have run. Maybe it's because of the recent github actions outage. I will close and open the pr to retrigger it. I had a similar situtation in one other PR and reopening pr worked.

srknzl:
Can we fix the github actions' not running issue? It is important to know that the workflows will run before merging the PR.

srknzl (commented):
Will this workflow work for PRs that originate from branches of the original repo? i.e A branch in hazelcast/hazelcast-python-client. We want workflow to work in this situation even though this is not how we develop.

srknzl (commented):
Will we be able to add this workflow as a required check when a PR comes from a community user? Looks like it will be decoupled from the PR which is not nice in my opinion.

Also a question: What happens when a PR comes from a community user? The workflow will run but will it fail ? I know run_tests job won't run, but does it mean a fail?

alisengul53:
If trigger mechanism is `pull_request`, it runs the workflow in the coming PR. But if the event trigger mechanism is `pull_request_target` it runs the workflow of the base branch which the PR is opened. For example in this case;

PR branch is `alisengul53:new_coverage_yaml`
Base branch is `hazelcast:master`

PR branch has a yaml including the trigger mechanism `pull_request_target`
Base branch has a yaml including the trigger mechanism `pull_request`

So in the PR I removed the old workflow which have the event trigger mechanism `pull_request`.
The new workflow has event trigger `pull_request_target` which can be runable when it is in the base branch which the PR is opened.

So it is normal to no workflow is trigger with this PR.

The best way to test this workflow before merging is to merge it to your forks then create a PR to your forked base branch.

.github/workflows/coverage_runner.yml
```diff
@@ -0,0 +1,109 @@
...
+          override_pr: ${{ github.event.inputs.pr_number }}
+          
+      - name: Upload remote controller logs if test run fails
+        uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: rc-logs-${{ matrix.python-version }}-${{ matrix.os }}
+          path: |
+            rc_stderr.log
+            rc_stdout.log
```
srknzl, line 109: Not a big issue we can make the newlines between steps consistent. Right now some of them has a newline some don't and I don't see a logical grouping either.

.github/workflows/coverage_runner.yml
```diff
@@ -0,0 +1,109 @@
...
+
+      - name: Checkout code for PR
+        if: github.event_name == 'pull_request_target'
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          
+      - name: Checkout repository to fetch PR coming from community
+        if: github.event_name == 'workflow_dispatch'
+        uses: actions/checkout@v2
```
srknzl, line 66: Why do we have three check out steps? The first and third can be merged. Can you explain why 2 is different?

.github/workflows/coverage_runner.yml
```diff
@@ -0,0 +1,109 @@
...
+        
+      - name: Fetch PR coming from community
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          git fetch origin pull/${{github.event.inputs.pr_number }}/head:temp_branch
```
srknzl, line 71: We can add a comment about `pull/${{github.event.inputs.pr_number }}/head`. It is not something widely known. I had to google it and found [this](https://stackoverflow.com/questions/63594658/git-refs-merge-vs-head-in-pull-request). I would add a comment like this:

"pull/<number>/head" is an automatically generated reference to HEAD commit of the PR.


Also can you rename `temp_branch` to pr_head or something like that to make it more clear?

alisengul53:
> Will we be able to add this workflow as a required check when a PR comes from a community user? Looks like it will be decoupled from the PR which is not nice in my opinion.
>

We can add this as a required check for community PRs, but first of all we need to review it and then if there is no security concern we can trigger it manually. If we are also happy with the result of the PR's coverage we can merge it. But these steps will not be handled automatically. So If we want, we can add this run as a required check.

> Also a question: What happens when a PR comes from a community user? The workflow will run but will it fail ? I know run_tests job won't run, but does it mean a fail?

The main idea is that if a PR is coming from community it shouldn't run automatically. The compleate workflow will be triggered but the actual test runner job will not be run. Generally job will seems as passed.

.github/workflows/coverage_runner.yml
```diff
@@ -0,0 +1,109 @@
...
+        env:
+          HAZELCAST_ENTERPRISE_KEY: ${{ secrets.HAZELCAST_ENTERPRISE_KEY }}
+        run: python run_tests.py
+            
+      - name: Publish results to Codecov for PR coming from hazelcast organization
+        if: ${{ matrix.python-version == '3.10' && matrix.os == 'ubuntu-latest' &&  github.event_name == 'pull_request_target' }}
+        uses: codecov/codecov-action@v1
+        with:
+          files: ./coverage.xml
+          override_pr: ${{ github.event.pull_request.number }}
```
srknzl, line 87: Maybe codecov action can determine the PR number itself. In the previous version of the file we were not specifying the PR number in `pull_request` event. They may have that support for `pull_request_target` as well.  If so we can merge two of these steps. This one and the one below.

.github/workflows/coverage_runner.yml
srknzl, line 109: Why did you deleted this part?

```
  run-linter:
    runs-on: ubuntu-latest
    name: Run black to check the code style
    steps:
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.10'
      - name: Checkout to code
        uses: actions/checkout@v2
      - name: Install dependencies
        run: |
          pip install -r requirements-dev.txt
      - name: Run black
        run: |
          black --check --config black.toml .
  generate_docs:
    runs-on: ubuntu-latest
    name: Generate documentation
    steps:
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.10'
      - name: Checkout to code
        uses: actions/checkout@v2
      - name: Install dependencies
        run: |
          pip install -r requirements-dev.txt
      - name: Generate documentation
        working-directory: docs
        run: |
          make html SPHINXOPTS="-W --keep-going -b linkcheck"
  run-mypy:
    runs-on: ubuntu-latest
    name: Run mypy to check type annotations
    steps:
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.10'
      - name: Checkout to code
        uses: actions/checkout@v2
      - name: Install dependencies
        run: |
          pip install -r requirements-dev.txt
      - name: Run mypy
        run: |
          mypy --show-error-codes hazelcast
```

srknzl (changes requested):

alisengul53:
> Will this workflow work for PRs that originate from branches of the original repo? i.e A branch in hazelcast/hazelcast-python-client. We want workflow to work in this situation even though this is not how we develop.

You can create a new branch from hazelcast/hazelcast-python-client and when you create a PR or any merge to `master` or `'[45].*.z'`  the workflow will be run

.github/workflows/coverage_runner.yml
alisengul53: Most probably when we remove `override_pr: ${{ github.event.pull_request.number }}` this line it will work but as the event trigger name is different, I feel more safe to specify PR number to the codecov action also. But If you are not happy with it, I can test it and if it works we can remove it. But I prefer to keep it.

.github/workflows/coverage_runner.yml
alisengul53: Fixed at [https://github.com/hazelcast/hazelcast-python-client/pull/539/commits/7cb06a7488d96c1f42d758b6ac12900be23a3051](url)

.github/workflows/coverage_runner.yml
alisengul53: I felt to separate steps for every event trigger would be more understandable but yes we can merge it and I merged it at https://github.com/hazelcast/hazelcast-python-client/pull/539/commits/7cb06a7488d96c1f42d758b6ac12900be23a3051
Second one checking out the PR automatically.
First and third one checking out the base branch. If it is a `workflow_dispatch` so if it is triggered by us, in the next step, a new temp `pr_head` branch will be fetched.

.github/workflows/coverage_runner.yml
alisengul53: Fixed at https://github.com/hazelcast/hazelcast-python-client/commit/7cb06a7488d96c1f42d758b6ac12900be23a3051

.github/workflows/coverage_runner.yml
alisengul53: As the new event trigger type is a little bit dangerous for security, I thought it will be good to use it when it is a need. For these jobs actually we don't need any secrets. I was planning to create new workflows for jobs which does not need any secrets. But you are right because if I am removing the old workflow, I should add the workflows I planned for these jobs.
Adding a new workflow for these jobs

srknzl:
> We can add this as a required check for community PRs, but first of all we need to review it and then if there is no security concern we can trigger it manually. If we are also happy with the result of the PR's coverage we can merge it. But these steps will not be handled automatically. So If we want, we can add this run as a required check.

Ok we can add it but how will it work? Let's say you added it as a required check and then go to actions tab of the repo and run the workflow. How will github match that run with the PR? I think in order not to forget running the workflow manually we need to make it mandatory somehow.

srknzl:
> The main idea is that if a PR is coming from community it shouldn't run automatically. The compleate workflow will be triggered but the actual test runner job will not be run. Generally job will seems as passed, but no coverage will be report to codecov

If it will be seen as passed, just making the workflow required won't do the trick because even if tests don't run the workflow will pass.

.github/workflows/coverage_runner.yml
srknzl: > First and third one checking out the base branch

Do not they check out the merge commit ? 🤔

.github/workflows/coverage_runner.yml
srknzl: Ok we can keep it, I was just curious.

.github/workflows/coverage_runner.yml
srknzl: > Second one checking out the PR automatically.

Ok as I see in https://github.com/actions/checkout, the default value for `ref` is the sha of the event:

> \# The branch, tag or SHA to checkout. When checking out the repository that
> \# triggered a workflow, this defaults to the reference or SHA for that event.
> \# Otherwise, uses the default branch.
> ref: ''

And you can see that the GITHUB_SHA of [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event is "Last commit on the PR base branch", and the GITHUB_SHA of [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) event is 
"Last merge commit on the GITHUB_REF branch". 

So I think we should try to checkout the merge commit again for the pull_request_target event. I think right now we are checking the pull_request head commit which is not the same thing. I think we should use ${{ github.event.pull_request.merge_commit_sha }}.

.github/workflows/coverage_runner.yml
```diff
@@ -0,0 +1,112 @@
...
+          java-version: '8'
+
+      - name: Checkout code for PR
+        if: github.event_name == 'pull_request_target'
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          
+      - name: Checkout repository for push event and PR coming from community
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event_name == 'push' }}
```
srknzl, line 65:
```suggestion
        if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'push' }}
```

srknzl (changes requested):

.github/workflows/coverage_runner.yml
alisengul53: >Do not they check out the merge commit ?
Checking out the merge commit and base branch is the same thing I think

>So I think we should try to checkout the merge commit again for the pull_request_target event
Why? We are trying to run test on the PR, why should we checkout the merge commit?

BTW I already tested these checks out on github action machines and AFAIR it was working as expected. But I will double check it

alisengul53:
> > We can add this as a required check for community PRs, but first of all we need to review it and then if there is no security concern we can trigger it manually. If we are also happy with the result of the PR's coverage we can merge it. But these steps will not be handled automatically. So If we want, we can add this run as a required check.
> 
> Ok we can add it but how will it work? Let's say you added it as a required check and then go to actions tab of the repo and run the workflow. How will github match that run with the PR? I think in order not to forget running the workflow manually we need to make it mandatory somehow.

There are not so much PR coming from community and for these PRs we must have a manuel process. I think it is not a big issue for us. We can check if we are able to add a check which can be triggered manually

.github/workflows/coverage_runner.yml
alisengul53: Thanks for catching it.

.github/workflows/coverage_runner.yml
srknzl: > Checking out the merge commit and base branch is the same thing I think

If checking out the base branch means checking out pull request's HEAD - I believe it does mean - the two are not the same thing. 

> Why? We are trying to run test on the PR, why should we checkout the merge commit?

Because that's the default behaviour of github. We may encounter with strange bugs. Github runs the tests on the merge commit because checkout action checks out to that commit. So if there are some changes in master that are not in the PR, those changes will be included in the test run as well. If you run it in the PR head, which is not the same thing as merge commit, we will have to merge master to pull request branch whenever master is updated. 

Workflows are running on merge commit because it proves the compatiblity when the changes of the PR is merged with the latest master. The merge commit is the new master commit after the merge which is reasonable to test.

alisengul53: I got your point.
I tried to handle it with `${{ github.event.pull_request.merge_commit_sha }}` but it checks out the latest commit of the base branch not merge commit. It didn't work as expected. Maybe it is related with `pull_request_target`
But [https://stackoverflow.com/questions/63594658/git-refs-merge-vs-head-in-pull-request#:~:text=The%20refs%2Fpull%2F%3Cnumber,as%20a%20%22future%22%20commit.](this) is what we need I think. I already tested it and it seems it is the thing we need. Lets have a demo togather to be sure

srknzl:
**Maybe we can do the following in the future (a note to future readers):**

Right now, the workflow will showed as passed after it is skipped if it comes from the community. In order not to forget triggering a manuel workflow before merging the PR we can do the following. The manual workflow dispatch can create some artifacts on the repo associated with a commit indicating if the workflow passed with that commit or not. The workflow in the PR would check the artifact and pass only if the manual workflow is passed with the latest commit of the pull request.


Since we don't have a lot of community PRs and implementing this seems a bit complex we decided to postpone this idea for now.

srknzl (approved):

PR #546, opened 2022-04-18
Updated Compact Configuration

yuce:
This PR:
* Changes `compact_serializers` configuration to be a list of `CompactSerializer`.
* Adds `get_class` method to the base `CompactSerializer`.
* Updates related docs and tests.
* Trivially updates `run_tests.py` to pass given comand line arguments to `pytest`.

### DIFF
--- a/docs/serialization.rst
+++ b/docs/serialization.rst
@@ -146,7 +146,7 @@ Then, a serializer for it can be implemented as below:
 
 .. code:: python
 
-    from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader
+    from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader, CompactSerializableClass
 
     class EmployeeSerializer(CompactSerializer[Employee]):
         def read(self, reader: CompactReader) -> Employee:
@@ -161,14 +161,17 @@ Then, a serializer for it can be implemented as below:
         def get_type_name(self) -> str:
             return "employee"
 
+        def get_class(self) -> CompactSerializableClass:
+            return Employee
+
 The last step is to register the serializer in the client configuration.
 
 .. code:: python
 
     client = HazelcastClient(
-        compact_serializers={
-            Employee: EmployeeSerializer(),
-        }
+        compact_serializers=[
+            EmployeeSerializer(),
+        ]
     )
 
 A schema will be created from the serializer, and a unique schema identifier
--- a/docs/serialization.rst
+++ b/docs/serialization.rst
@@ -1,3 +1,4 @@
+
 Serialization
 =============
 
@@ -149,19 +150,19 @@ Then, a serializer for it can be implemented as below:
     from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader, CompactSerializableClass
 
     class EmployeeSerializer(CompactSerializer[Employee]):
-        def read(self, reader: CompactReader) -> Employee:
+        def read(self, reader: CompactReader):
             name = reader.read_string("name")
             age = reader.read_int32("age")
             return Employee(name, age)
 
-        def write(self, writer: CompactWriter, obj: Employee) -> None:
+        def write(self, writer: CompactWriter, obj: Employee):
             writer.write_string("name", obj.name)
             writer.write_int32("age", obj.age)
 
-        def get_type_name(self) -> str:
+        def get_type_name(self):
             return "employee"
 
-        def get_class(self) -> CompactSerializableClass:
+        def get_class(self):
             return Employee
 
 The last step is to register the serializer in the client configuration.
--- a/docs/serialization.rst
+++ b/docs/serialization.rst
@@ -146,7 +146,7 @@ Then, a serializer for it can be implemented as below:
 
 .. code:: python
 
-    from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader, CompactSerializableClass
+    from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader
 
     class EmployeeSerializer(CompactSerializer[Employee]):
         def read(self, reader: CompactReader):
--- a/docs/serialization.rst
+++ b/docs/serialization.rst
@@ -1,4 +1,3 @@
-
 Serialization
 =============
 
--- a/examples/serialization/compact_serialization_example.py
+++ b/examples/serialization/compact_serialization_example.py
@@ -1,5 +1,10 @@
 from hazelcast import HazelcastClient
-from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader, CompactSerializableClass
+from hazelcast.serialization.api import (
+    CompactSerializer,
+    CompactWriter,
+    CompactReader,
+    CompactSerializableClass,
+)
 
 
 class Address:
@@ -57,9 +62,7 @@ class EmployeeSerializer(CompactSerializer[Employee]):
         return Employee
 
 
-client = HazelcastClient(
-    compact_serializers=[AddressSerializer(), EmployeeSerializer()]
-)
+client = HazelcastClient(compact_serializers=[AddressSerializer(), EmployeeSerializer()])
 
 employees = client.get_map("employees").blocking()
 
--- a/examples/serialization/compact_serialization_example.py
+++ b/examples/serialization/compact_serialization_example.py
@@ -3,7 +3,6 @@ from hazelcast.serialization.api import (
     CompactSerializer,
     CompactWriter,
     CompactReader,
-    CompactSerializableClass,
 )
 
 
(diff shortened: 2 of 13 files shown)


hazelcast/config.py
```diff
@@ -1369,19 +1369,15 @@ def use_public_ip(self, value):
...
-    def compact_serializers(self) -> typing.Dict[typing.Type, CompactSerializer]:
+    def compact_serializers(self) -> typing.List[CompactSerializer]:
         return self._compact_serializers
 
     @compact_serializers.setter
-    def compact_serializers(self, value: typing.Dict[typing.Type, CompactSerializer]) -> None:
-        if isinstance(value, dict):
-            for clazz, serializer in value.items():
-                if not isinstance(clazz, type):
-                    raise TypeError("Keys of compact_serializers must be classes")
```
srknzl, line 1380: We are losing the validation of classes here. Can we somehow check that getClass method exists in the serializer and it returns a class and nothing else. Does instanceof check takes care of the existence of methods ? I am sure it does not  handle the validation of the return values of getClass and getTypeName :). 

I know the `get_type_name` validation was nonexistent before as well but I think we can add it. In Node.js, because of fail fast goal of ours, I am calling the methods and checking the return value's type. See https://github.com/hazelcast/hazelcast-nodejs-client/blob/801696baf6492475ad1ddf664356a9707a0bf092/src/config/ConfigBuilder.ts#L512 and https://github.com/hazelcast/hazelcast-nodejs-client/blob/801696baf6492475ad1ddf664356a9707a0bf092/src/config/ConfigBuilder.ts#L517

docs/serialization.rst
```diff
@@ -146,7 +146,7 @@ Then, a serializer for it can be implemented as below:
 
 .. code:: python
 
-    from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader
+    from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader, CompactSerializableClass
```
srknzl, line 149: Don't forget to change this file if you change `get_class` method's return type.

examples/serialization/compact_serialization_example.py
```diff
@@ -1,5 +1,10 @@
 from hazelcast import HazelcastClient
-from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader
+from hazelcast.serialization.api import (
+    CompactSerializer,
+    CompactWriter,
+    CompactReader,
+    CompactSerializableClass,
```
srknzl, line 6: Don't forget to change this file if you change `get_class` method's return type.

examples/sql/sql_compact_example.py
```diff
@@ -1,7 +1,12 @@
 import dataclasses
 
 from hazelcast import HazelcastClient
-from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader
+from hazelcast.serialization.api import (
```
srknzl, line 4: Don't forget to change this file if you change `get_class` method's return type.

hazelcast/serialization/service.py
```diff
@@ -344,7 +344,7 @@ def __init__(
         self._registration_lock = threading.RLock()
         self._int_type_id = _int_type_to_type_id.get(config.default_int_type, None)
 
-        self._compact_types = set(config.compact_serializers.keys())
+        self._compact_types = [c.get_class() for c in config.compact_serializers]
```
srknzl, line 347: Does not using a dict make more sense here? We will do a lot of lookups in this data type.

run_tests.py
```diff
@@ -21,29 +21,31 @@ def wait_until_rc_is_ready():
...
-
-                process = subprocess.run(args)
-                rc_process.kill()
-                rc_process.wait()
-                sys.exit(process.returncode)
-            except:
-                rc_process.kill()
-                rc_process.wait()
-                raise
+with open("rc_stdout.log", "w") as rc_stdout:
```
srknzl, line 24: I think you did that becuase there is no files that import run_tests.py. Are we sure there won't be in the future? It is safe to include a guard `if __name__ == "__main__":`

run_tests.py
```diff
@@ -21,29 +21,31 @@ def wait_until_rc_is_ready():
...
+        try:
+            wait_until_rc_is_ready()
+            sys.stdout.flush()
+
+            args = [
+                "pytest",
+                "--cov=hazelcast",
+                "--cov-report=xml",
+            ]
+            args.extend(sys.argv[1:])
```
srknzl, line 36: Why did you add this line?

srknzl (changes requested):

run_tests.py
yuce: I don't think we'll use this file as a library.

run_tests.py
yuce: To make arguments passed to `run_tests` passed to pytest, like:
```
./run_tests.py -k FilterMyTest
```

hazelcast/serialization/service.py
yuce: It should be a set instead of a list. Will update.

srknzl: It is even better (O(1)) to make it a dict than a set (O(log(n)))

yuce: Are you sure set lookup is O(log(n))? According to https://wiki.python.org/moin/TimeComplexity it is O(1).

srknzl: Ok I was thinking it is like a set in cpp.

hazelcast/serialization/api.py
```diff
@@ -3059,6 +3059,14 @@ def write(self, writer: CompactWriter, obj: CompactSerializableClass) -> None:
                 to write.
         """
 
+    @abc.abstractmethod
+    def get_class(self) -> CompactSerializableClass:
+        """Returns the class for this serializer.
+
+        Returns:
+            The class for this serializer.
```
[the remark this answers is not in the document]
yuce: Other docs for this class doesn't have examples, so I've left that out. I've made the method's purpose clearer at: https://github.com/hazelcast/hazelcast-python-client/pull/546/commits/ae2532d913aad0f7b02f20cacd0852209ea024c4

hazelcast/config.py
yuce: Since the user has to derive their class from `CompactSerializer)`, the class will have its `get_class` method.

docs/serialization.rst
yuce: Removed the import for CompactSerializableClass at: https://github.com/hazelcast/hazelcast-python-client/pull/546/commits/ae2532d913aad0f7b02f20cacd0852209ea024c4

examples/serialization/compact_serialization_example.py
yuce: Removed the import for that at: https://github.com/hazelcast/hazelcast-python-client/pull/546/commits/6f6d8fdd8681b6349a84526961a55ef96e1a4eed

examples/sql/sql_compact_example.py
yuce: At: https://github.com/hazelcast/hazelcast-python-client/pull/546/commits/6f6d8fdd8681b6349a84526961a55ef96e1a4eed

docs/serialization.rst
srknzl: Looks like you have not removed the import yet.

```suggestion
    from hazelcast.serialization.api import CompactSerializer, CompactWriter, CompactReader
```

hazelcast/config.py
srknzl: What about return value validation?

yuce: That's something to be caught by the type checker.

docs/serialization.rst
yuce: My comments keep disappearing. Removed that at: https://github.com/hazelcast/hazelcast-python-client/pull/546/commits/df32f6dae22693344c393081f1a7339517051218

hazelcast/config.py
srknzl: However not everyone uses a type checker in python. 

Think of Node.js client. For non-ts users we do a runtime check even we don't have to. Isn't this what we want to do, failing fast? 

In java, we the compilation step validates the type. 

If we have the goal of failing fast given a  wrong configuration, we should do validate here.

srknzl: We will discuss later how much to validate. Skipping now.

srknzl (approved):

docs/serialization.rst
```diff
@@ -149,26 +149,29 @@ Then, a serializer for it can be implemented as below:
...
+
 The last step is to register the serializer in the client configuration.
 
 .. code:: python
 
     client = HazelcastClient(
-        compact_serializers={
-            Employee: EmployeeSerializer(),
-        }
+        compact_serializers=[
```
utku-caglayan, line 172: Can you update the documentation for this parameter? It looks like user should pass a dictionary, based on that.

utku-caglayan (dismissed):

docs/serialization.rst
yuce: No, the user passes a list of serializers.

srknzl (approved):

utku-caglayan (approved):

docs/serialization.rst
yuce: Thanks for catching the outdated documentation. Fixed at: https://github.com/hazelcast/hazelcast-python-client/pull/546/commits/878d5ed0a636f79ab81034483582c1767837806d

srknzl (approved):

utku-caglayan (approved):

PR #557, opened 2022-07-03
Cloud APIs are added and auto-generated files are updated

alisengul53:


yuce:
@alisengul53 linter task fails because the "black" formatter doesn't like some code's style. You can resolve that by installing `black`:
```
pip install black
```
And then running the following in the project root:
```
black --config black.toml .
```

PR #567, opened 2022-08-19
Update Compact related tests for backward compatibility tests

mdumandag:
We need to skip 5.2 server versions if the client version is <5.2.

This PR updates the tests to skip on such scenarious.


srknzl (approved):

mdumandag:
Merging with one approval since it is a small change

PR #571, opened 2022-09-06
Support for running tests with PyPy3

mehmettokgoz:


hazelcast/near_cache.py
```diff
@@ -171,6 +172,12 @@ def __getitem__(self, key):
             else value_record.value
         )
 
+    def _get_size_of_near_cache(self):
+        if "PyPy" in sys.version:
+            return 0
+        else:
+            return getsizeof(self)
+
```
yuce, line 180: `getsizeof` can take an argument for the default, I think we can just use it instead of checking for the PyPy version:
```
"owned_entry_memory_cost": getsizeof(self, 0),
```

mehmettokgoz: Fixed at https://github.com/hazelcast/hazelcast-python-client/pull/571/commits/3f20f47c2540ebb4248f33ffd8deb8e977962a55.

.github/workflows/nightly_runner.yml
```diff
@@ -9,12 +9,12 @@ jobs:
     name: Run tests with Python ${{ matrix.python-version }} on ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [ '3.6', '3.7', '3.8', '3.9', '3.10' ]
+        python-version: [ '3.6', '3.7', '3.8', '3.9', '3.10', 'pypy-3.7', 'pypy-3.8', 'pypy-3.9']
```
yuce, line 12: IMO it's adequate to check only the latest PyPy version.

mehmettokgoz:
FYI, I ran the tests and results are here https://github.com/mehmettokgoz/hazelcast-python-client/actions/runs/3090554387/jobs/5010540870.

.github/workflows/nightly_runner.yml
mehmettokgoz: Fixed at https://github.com/hazelcast/hazelcast-python-client/pull/571/commits/2877cd06a3efb7abd4ec357a3842a40c80b74edc.

yuce (approved):

Issue #556, opened 2022-06-24
Potential compact bug, see description  [API-1416]


srknzl:
The same bug in https://github.com/hazelcast/hazelcast-nodejs-client/issues/1323 seems to exist in python client as well.

PR #574, opened 2022-09-12
Nested compact field's serializer can be None if it is not registered [API-1416]

mehmettokgoz:
Fixes https://github.com/hazelcast/hazelcast-python-client/issues/556

Added check for None serializer, on it, throw a HazelcastSerializationError in this case.
Added a new test for missing nested serializer case, it checks if a HazelcastSerializationError has been thrown.

### DIFF
--- a/hazelcast/serialization/compact.py
+++ b/hazelcast/serialization/compact.py
@@ -46,7 +46,7 @@ class CompactStreamSerializer(BaseSerializer):
         serializer = self._type_to_serializer.get(clazz)
         if serializer is None:
             raise HazelcastSerializationError(
-                "No serializer is registered for class {}.".format(type(obj).__name__)
+                f"No serializer is registered for class/constructor {clazz.__name__}."
             )
         schema = self._type_to_schema.get(clazz)
         if not schema:
--- a/hazelcast/serialization/compact.py
+++ b/hazelcast/serialization/compact.py
@@ -45,9 +45,7 @@ class CompactStreamSerializer(BaseSerializer):
         clazz = type(obj)
         serializer = self._type_to_serializer.get(clazz)
         if serializer is None:
-            raise HazelcastSerializationError(
-                f"No serializer is registered for class {clazz}."
-            )
+            raise HazelcastSerializationError(f"No serializer is registered for class {clazz}.")
         schema = self._type_to_schema.get(clazz)
         if not schema:
             schema = CompactStreamSerializer._build_schema(serializer, obj)
--- a/hazelcast/serialization/compact.py
+++ b/hazelcast/serialization/compact.py
@@ -46,7 +46,7 @@ class CompactStreamSerializer(BaseSerializer):
         serializer = self._type_to_serializer.get(clazz)
         if serializer is None:
             raise HazelcastSerializationError(
-                f"No serializer is registered for class/constructor {clazz}."
+                f"No serializer is registered for class {clazz}."
             )
         schema = self._type_to_schema.get(clazz)
         if not schema:
--- a/hazelcast/serialization/compact.py
+++ b/hazelcast/serialization/compact.py
@@ -43,7 +43,9 @@ class CompactStreamSerializer(BaseSerializer):
 
     def write(self, out: ObjectDataOutput, obj: typing.Any) -> None:
         clazz = type(obj)
-        serializer = self._type_to_serializer[clazz]
+        serializer = self._type_to_serializer.get(clazz)
+        if serializer is None:
+            raise HazelcastSerializationError('No serializer is registered for class {}.'.format(type(obj).__name__))
         schema = self._type_to_schema.get(clazz)
         if not schema:
             schema = CompactStreamSerializer._build_schema(serializer, obj)
--- a/hazelcast/serialization/compact.py
+++ b/hazelcast/serialization/compact.py
@@ -46,7 +46,7 @@ class CompactStreamSerializer(BaseSerializer):
         serializer = self._type_to_serializer.get(clazz)
         if serializer is None:
             raise HazelcastSerializationError(
-                f"No serializer is registered for class/constructor {clazz.__name__}."
+                f"No serializer is registered for class/constructor {clazz}."
             )
         schema = self._type_to_schema.get(clazz)
         if not schema:
--- a/hazelcast/serialization/compact.py
+++ b/hazelcast/serialization/compact.py
@@ -45,7 +45,9 @@ class CompactStreamSerializer(BaseSerializer):
         clazz = type(obj)
         serializer = self._type_to_serializer.get(clazz)
         if serializer is None:
-            raise HazelcastSerializationError('No serializer is registered for class {}.'.format(type(obj).__name__))
+            raise HazelcastSerializationError(
+                "No serializer is registered for class {}.".format(type(obj).__name__)
+            )
         schema = self._type_to_schema.get(clazz)
         if not schema:
             schema = CompactStreamSerializer._build_schema(serializer, obj)
--- a/tests/integration/backward_compatible/serialization/compact_test.py
+++ b/tests/integration/backward_compatible/serialization/compact_test.py
@@ -784,6 +784,26 @@ class CompactWithListenerTest(CompactTestBase):
         self.assertTrueEventually(lambda: self.assertEqual(1, counter.get()))
 
 
+class CompactWithNestedFieldTest(CompactTestBase):
+
+    def test_missing_nested_serializer(self):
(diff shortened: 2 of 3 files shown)


hazelcast/serialization/compact.py
```diff
@@ -43,7 +43,11 @@ def __init__(self, compact_serializers: typing.List[CompactSerializer]):
 
     def write(self, out: ObjectDataOutput, obj: typing.Any) -> None:
         clazz = type(obj)
-        serializer = self._type_to_serializer[clazz]
+        serializer = self._type_to_serializer.get(clazz)
+        if serializer is None:
+            raise HazelcastSerializationError(
+                "No serializer is registered for class {}.".format(type(obj).__name__)
```
mdumandag, line 49:
```suggestion
                f"No serializer is registered for class {clazz}."
```

Let's convert it to this to make use of f-strings and printing the class name with the package, which will be more useful for debugging

tests/integration/backward_compatible/serialization/compact_test.py
```diff
@@ -784,6 +784,24 @@ def listener(_):
         self.assertTrueEventually(lambda: self.assertEqual(1, counter.get()))
 
 
+class CompactWithNestedFieldTest(CompactTestBase):
```
mdumandag, line 787: I feel like this can easily be a unit test. Can you move it to `tests/unit/serialization/compact_test.py` and modify it such that it doesn't need a cluster? You can take a look at the other tests in that directory for reference. Just construct a serialization service with appropriate configuration, call to_data on it and expect it to throw the exception

mehmettokgoz: Converted it to a unit test at https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/3e052df6aaaf866e1f977a76832ca4c1577e85f4. I had to a custom schema registration function to simulate the `put()` function of `MapProxy`. Node.js Client has similar util functions too. See at [CompactUtil.js#L1483](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/test/integration/backward_compatible/parallel/serialization/compact/CompactUtil.js#L1483).

hazelcast/serialization/compact.py
mehmettokgoz: Refactored at https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/3e052df6aaaf866e1f977a76832ca4c1577e85f4. Using `clazz` without `__name__` gives output as follows: `No serializer is registered for class/constructor <class 'tests.integration.backward_compatible.serialization.compact_test.Nested'>.`. </br></br> I added `__name__` to sync the output with Node.js client, should we use the long output, wdyt @mdumandag?

mdumandag: I believe we should use the long output, because it is easier to understand what class we are talking about, when it also has the package name.

tests/unit/serialization/compact_test.py
```diff
@@ -200,3 +209,41 @@ def test_schema_writer(self):
...
+    def _register_schema(self, service, obj):
+        writer = SchemaWriter(type(obj).__name__)
+        fields = []
+        for kind in FIELD_KINDS:
+            name = kind.name.lower()
+            fun = getattr(writer, f"write_{kind.name.lower()}", None)
+            if fun is None:
+                continue
+            fun(name, None)
+            fields.append((name, kind))
```
mdumandag, line 224: This is just `serializer.write(writer, obj)`, right? If so, can we make use of it to simplify this?

tests/unit/serialization/compact_test.py
```diff
@@ -200,3 +209,41 @@ def test_schema_writer(self):
...
+        service.compact_stream_serializer.register_schema_to_type(schema, type(obj))
+
+    def _serialize(self, serializationService, obj):
+        try:
+            return serializationService.to_data(obj)
+        except SchemaNotReplicatedError:
+            self._register_schema(serializationService, obj)
+            return self._serialize(serializationService, obj)
+        except:
+            raise
```
mdumandag, line 235: No need to do this, if the catched exception is not `SchemaNotReplicatedError`, it will throw anyways

tests/unit/serialization/compact_test.py
```diff
@@ -200,3 +209,41 @@ def test_schema_writer(self):
...
+            name = kind.name.lower()
+            fun = getattr(writer, f"write_{kind.name.lower()}", None)
+            if fun is None:
+                continue
+            fun(name, None)
+            fields.append((name, kind))
+        schema = writer.build()
+        service.compact_stream_serializer.register_schema_to_type(schema, type(obj))
+
+    def _serialize(self, serializationService, obj):
```
mdumandag, line 228:
```suggestion
    def _serialize(self, serialization_service, obj):
```
we should use camel case

tests/unit/serialization/compact_test.py
```diff
@@ -200,3 +209,41 @@ def test_schema_writer(self):
...
+        try:
+            return serializationService.to_data(obj)
+        except SchemaNotReplicatedError:
+            self._register_schema(serializationService, obj)
+            return self._serialize(serializationService, obj)
+        except:
+            raise
+
+    def test_missing_serializer(self):
+        serializer = SomeFieldsSerializer.from_kinds(FIELD_KINDS)
```
mdumandag, line 238: I believe instead of using this big class with lots of fields, we can define two very simple classes in this test class, and make use them. It might be hard to understand at first why this test fails. I know it fails because we don't have the serializer for `Nested`, but it is not obvious for future maintainers from this file

mehmettokgoz: I have defined a Parent and Child class and replaced them with the `SomeField` type. See: https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/7dc27039f9764d3aa88134a3885745041d84b096

tests/unit/serialization/compact_test.py
mehmettokgoz: Refactored at https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/7dc27039f9764d3aa88134a3885745041d84b096.

tests/unit/serialization/compact_test.py
mehmettokgoz: Removed at https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/7dc27039f9764d3aa88134a3885745041d84b096.

tests/unit/serialization/compact_test.py
mehmettokgoz: This logic is replaced with a simple schema put call, see https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/7dc27039f9764d3aa88134a3885745041d84b096.

hazelcast/serialization/compact.py
```diff
@@ -43,7 +43,11 @@ def __init__(self, compact_serializers: typing.List[CompactSerializer]):
 
     def write(self, out: ObjectDataOutput, obj: typing.Any) -> None:
         clazz = type(obj)
-        serializer = self._type_to_serializer[clazz]
+        serializer = self._type_to_serializer.get(clazz)
+        if serializer is None:
+            raise HazelcastSerializationError(
+                f"No serializer is registered for class/constructor {clazz}."
```
mdumandag, line 49:
```suggestion
                f"No serializer is registered for class {clazz}."
```
There is no constructor for Python

tests/unit/serialization/compact_test.py
```diff
@@ -14,6 +17,13 @@
     FIELD_OPERATIONS,
     _BOOLEANS_PER_BYTE,
     FieldKind,
+    SchemaNotReplicatedError,
+)
+from tests.integration.backward_compatible.serialization.compact_test import (
+    SomeFieldsSerializer,
+    FIELD_KINDS,
+    REFERENCE_OBJECTS,
```
mdumandag, line 25: I assume some of these imports are now unused

mdumandag (commented):
LGTM with the recent changes. Just left two more minor comments. Then, we can merge the PR

tests/unit/serialization/compact_test.py
mehmettokgoz: Removed at https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/53b1b2f5f916d56afa44207f68d6951697840359.

hazelcast/serialization/compact.py
mehmettokgoz: Removed at https://github.com/hazelcast/hazelcast-python-client/pull/574/commits/53b1b2f5f916d56afa44207f68d6951697840359.

mdumandag (approved):

PR #586, opened 2022-09-27
DBAPI Support [API-911]

yuce:
- Adds DBAPI 2 compatiblity layer. The documentation and code samples will be comitted in a separate PR.
- Adds a Makefile to do various operations easier.

The standard DBAPI 2 test passes without changes.


hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+            self._closed = True
+            self._conn._close_cursor(self)
+            if self._res:
+                self._res.close()
+                self._res = None
+
+    def execute(self, operation: str, params: Optional[Tuple] = None) -> None:
+        if params is not None and not isinstance(params, tuple):
+            raise InterfaceError("params must be a tuple or None")
+        params = params or ()
```
mdumandag, line 139: why do we need this? at this point, we know it is a non-null tuple

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+        futures = []
+        svc = self._conn._get_client().sql
+        for params in seq_of_params:
+            futures.append(
+                svc.execute(
+                    operation, *params, expected_result_type=SqlExpectedResultType.UPDATE_COUNT
+                )
+            )
+        for fut in futures:
+            fut.result()
```
mdumandag, line 173: what happens after this line? after all the operations are completed, we are not setting anything in the fields. How do we use the results(update counts)?

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+
+    def fetchone(self) -> Optional[SqlRow]:
+        if self._iter is None:
+            raise InterfaceError("fetch can only be called after row returning queries")
+        if self._rownumber is None:
```
mdumandag, line 178: when can row_number be none?

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+            row = next(self._iter)
+            self._rownumber += 1
+            return row
+        except StopIteration:
+            return None
+
+    def fetchmany(self, size: Optional[int] = None) -> List[SqlRow]:
+        if self._iter is None:
+            raise InterfaceError("fetchmany can only be called after row returning queries")
+        if self._rownumber is None:
```
mdumandag, line 190: when can row_number be none?

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+        if size is None:
+            size = self.arraysize
+        rows = list(itertools.islice(self._iter, size))
+        self._rownumber += len(rows)
+        return rows
+
+    def fetchall(self) -> List[SqlRow]:
+        if self._iter is None:
+            raise InterfaceError("fetchall can only be called after row returning queries")
+        if self._rownumber is None:
```
mdumandag, line 201: when can row_number be none?

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+            setattr(cfg, attr_name, value)
+        else:
+            raise InterfaceError(f"Unknown DSN attribute: {k}")
+    return cfg
+
+
+def _make_bool(v: str) -> bool:
+    if v == "true":
+        return True
+    if v == "false":
```
mdumandag, line 495: Should we also allow capitalized `True` and `False`?

tests/integration/backward_compatible/db_test.py
```diff
@@ -0,0 +1,123 @@
...
+    SqlTestBase,
+    compare_server_version_with_rc,
+    compare_client_version,
+    SERVER_CONFIG,
+    JET_ENABLED_CONFIG,
+    Student,
+)
+
+
+class DbapiTestBase(SqlTestBase):
```
mdumandag, line 16: The test should be guarded against the client and server version

tests/integration/backward_compatible/dbapi20.py
```diff
@@ -0,0 +1,831 @@
...
+    -- Ian Bicking
+"""
+
+__version__ = "1.15.0"
+
+import unittest
+import time
+import sys
+
+if sys.version[0] >= "3":  # python 3.x
```
mdumandag, line 20: do we need those?

tests/integration/backward_compatible/dbapi20.py
```diff
@@ -0,0 +1,831 @@
+#!/usr/bin/env python
```
mdumandag, line 1: We need to add server and client version guards to this test (since it is in the backward compatible folder)

tests/integration/backward_compatible/dbapi20.py
```diff
@@ -0,0 +1,831 @@
...
+        cursor.execute(self.ddl1)
+
+    def executeDDL2(self, cursor):
+        cursor.execute(self.ddl2)
+
+    def setUp(self):
+        """self.drivers should override this method to perform required setup
+        if any is necessary, such as creating the database.
+        """
+        pass
```
mdumandag, line 88: how is this test working with no setup on our side?

tests/integration/backward_compatible/hazelcast_dbapi20_test.py
```diff
@@ -0,0 +1,61 @@
+from hazelcast.config import Config
+import tests.integration.backward_compatible.dbapi20 as dbapi20
+from hazelcast import db
+from .sql_test import SqlTestBase
+
+
+class test_HazelcastDBAPI20(SqlTestBase, dbapi20.DatabaseAPI20Test):
```
mdumandag, line 7: Can we use a more conventional name for the class?

tests/unit/dbapi_test.py
```diff
@@ -0,0 +1,133 @@
+import unittest
+
+from hazelcast.config import Config
+from hazelcast.db import _make_config, InterfaceError
+
+
+class DbApiTestCase(unittest.TestCase):
```
mdumandag, line 7:
```suggestion
class DbApiTest(unittest.TestCase):
```

to be consistent with our other tests

tests/unit/dbapi_test.py
```diff
@@ -0,0 +1,133 @@
...
+        target = config_with_values(cluster_members=["localhost:1234"])
+        self.assertEqualConfig(target, cfg)
+
+    def test_make_config_host_port(self):
+        cfg = _make_config(host="foo.com", port=1234)
+        target = config_with_values(cluster_members=["foo.com:1234"])
+        self.assertEqualConfig(target, cfg)
+
+    def test_make_config_dsn(self):
+        test_cases = [
```
mdumandag, line 60: we are using parameterized tests in the python client for such things

tests/unit/dbapi_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+def config_with_values(**kwargs) -> Config:
+    return Config.from_dict(kwargs)
+
+
+def config_to_dict(cfg: Config) -> dict:
+    d = {}
+    for k in cfg.__slots__:
+        d[k] = getattr(cfg, k)
+    return d
```
mdumandag, line 133: Can you add a test that verifies dsn is parsed correctly when there are problematic characters in it? For example, parsing things like `hz://user?smart=True:pass@foo.com:1234`

tests/integration/backward_compatible/db_test.py
```diff
@@ -0,0 +1,123 @@
...
+        result = list(c.fetchall())
+        self.assertCountEqual(
+            [(i, i) for i in range(5)],
+            [(row.get_object("__key"), row.get_object("this")) for row in result],
+        )
+        self.assertEqual(5, c.rownumber)
+
+    def test_cursor_connection(self):
+        c = self.conn.cursor()
+        self.assertEqual(self.conn, c.connection)
```
mdumandag, line 123: can you add a test for description, as we return our own class there?

tests/integration/backward_compatible/db_test.py
```diff
@@ -0,0 +1,123 @@
+from typing import List
```
mdumandag, line 1: I believe if we say that we provide thread safety for module and connection, we should add a test for it

hazelcast/db.py
yuce: `params` may be None.

hazelcast/db.py
yuce: Hz SQL always returns `0` for update count, so the result is not important.

hazelcast/db.py
yuce: This is just the DSN, so only `true` and `false` should be sufficient.

tests/integration/backward_compatible/dbapi20.py
yuce: I've copied that file verbatim from the original source. I think we can keep them unless they are causing issues.

tests/unit/dbapi_test.py
yuce: I think the current approach is simpler and easier to understand.

tests/unit/dbapi_test.py
yuce: Which is the problematic character in that example?
We use the standard `urllib.parse.urlparse` function, so anything parsed by that function should be OK.

tests/integration/backward_compatible/db_test.py
yuce: `DescriptionColumn` is not a class but a named tuple, so equivalent to the tuple in the spec. but will add a test.

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+        "display_size",
+        "internal_size",
+        "precision",
+        "scale",
+        "null_ok",
+    ],
+)
+
+
+class Type(enum.Enum):
```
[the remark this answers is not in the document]
mdumandag: I see, now as I read more of the specification, it says that 
```
A [Cursor](https://peps.python.org/pep-0249/#id12) Object’s description attribute returns information about each of the result columns of a query. The type_code must compare equal to one of Type Objects defined below. Type Objects may be equal to more than one type code (e.g. DATETIME could be equal to the type codes for date, time and timestamp columns; see the [Implementation Hints](https://peps.python.org/pep-0249/#implementation-hints) below for details).
```

so the types we have here must match the ones defined below. we can't probably define types like decimal, or JSON according to this paragraph

hazelcast/db.py
mdumandag: OK,  I missed the `params is not None` part

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+    _DEFAULT_CURSOR_BUFFER_SIZE,
+)
+
+apilevel = "2.0"
+# Threads may share the module and connections.
+threadsafety = 2
+paramstyle = "qmark"
+
+DescriptionColumn = namedtuple(
+    "DescriptionColumn",
```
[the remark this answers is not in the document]
yuce: Yes, it's the description of a column in a row.

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+
+class RowResult:
+    def __init__(self, result):
+        self._result = result
+
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
```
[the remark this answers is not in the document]
yuce: Removed at: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/6f0929e54cd6bac8cc755b1bae86fc73245df4a1

hazelcast/db.py
yuce: Leftover, removed at: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/6f0929e54cd6bac8cc755b1bae86fc73245df4a1

hazelcast/db.py
yuce: Leftover, removed at: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/6f0929e54cd6bac8cc755b1bae86fc73245df4a1

hazelcast/db.py
yuce: Leftover, removed at: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/6f0929e54cd6bac8cc755b1bae86fc73245df4a1

tests/integration/backward_compatible/db_test.py
```diff
@@ -0,0 +1,123 @@
...
+
+from hazelcast import HazelcastClient
+from hazelcast.config import Config
+from hazelcast.db import connect, Connection
```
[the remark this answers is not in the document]
yuce: Moved these tests out out of backward compatibility.

tests/integration/backward_compatible/dbapi20.py
yuce: Moved these tests out out of backward compatibility.

tests/integration/backward_compatible/db_test.py
yuce: Moved these tests out out of backward compatibility.

tests/integration/backward_compatible/hazelcast_dbapi20_test.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/6f0929e54cd6bac8cc755b1bae86fc73245df4a1

tests/unit/dbapi_test.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/6f0929e54cd6bac8cc755b1bae86fc73245df4a1

tests/unit/dbapi_test.py
mdumandag: Maybe simpler to write but when things go wrong, they are not good.
- We can't run all the test cases once the first one fails
- We can't know which test case failed directly from the output

tests/unit/dbapi_test.py
mdumandag: I mean having characters like `:`, `?`, `=` in the values can result in weird parsing if they are not handled correctly

hazelcast/db.py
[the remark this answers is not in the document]
mdumandag: Then, should we rename it to `ColumnDescription`?

tests/integration/backward_compatible/db_test.py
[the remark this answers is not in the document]
mdumandag: When I think more about it, the backward compatibility tests are for testing the compatibility of the server. And, this is basically a client-only feature on top of the backward compatible SQL feature. So, I think it might be OK to leave it like this. But, we still need to add server version checks to these tests to make sure that they will run correctly in client backward compatibility tests

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+    DECIMAL = 8
+    JSON = 9
+    OBJECT = 10
+
+
+Date = date
+Time = time
+Timestamp = datetime
+Binary = bytes
+STRING = Type.STRING
```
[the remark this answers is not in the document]
yuce: You are right. That was one of the things I couldn't decide about the API, but I've fixed this using a similar solution with pyscopg2 tand MySQL drivers at: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/1094643ff66716fc22b654629e88ad4a9101c64a

hazelcast/__init__.py
```diff
@@ -7,3 +7,4 @@
 logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 from hazelcast.client import HazelcastClient
+from hazelcast.db import connect
```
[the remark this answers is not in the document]
yuce: That makes sense: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/41a3be8afdc5bf39cdeac88cc0a34fdc3210be66

hazelcast/db.py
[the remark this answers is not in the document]
yuce: OK: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/41a3be8afdc5bf39cdeac88cc0a34fdc3210be66

tests/unit/dbapi_test.py
yuce: The current way of implementing this thing in the code base looks clunky and very limited to me. Using subtests would be better, but it requires a depencency for pytest: https://pypi.org/project/pytest-subtests/ 

> We can't run all the test cases once the first one fails

Is it that important? It takes a second to run the tests again if one of them fails.

> We can't know which test case failed directly from the output

I've improved the tests to display a message about the test case at: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/41a3be8afdc5bf39cdeac88cc0a34fdc3210be66

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+    SqlExpectedResultType,
+    _DEFAULT_CURSOR_BUFFER_SIZE,
+)
+
+apilevel = "2.0"
+# Threads may share the module and connections.
+threadsafety = 2
+paramstyle = "qmark"
+
+DescriptionColumn = namedtuple(
```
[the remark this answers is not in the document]
yuce: I'll submit the documentation and code samples in another PR.

tests/integration/backward_compatible/db_test.py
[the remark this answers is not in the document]
yuce: The server version is already checked, since the tests inherit "SqlTestBase"

hazelcast/db.py
```diff
@@ -0,0 +1,509 @@
...
+    "DescriptionColumn",
+    [
+        "name",
+        "type",
+        "display_size",
+        "internal_size",
+        "precision",
+        "scale",
+        "null_ok",
```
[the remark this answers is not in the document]
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/d4e25c7e39dfe7016bd6237b3a6ad8e7112e7ded

tests/integration/backward_compatible/db_test.py
yuce: Added a test which creates cursors in parallel at: https://github.com/hazelcast/hazelcast-python-client/pull/586/commits/97482355eb0465ec4b0543b5dc1f48df0800db88

mdumandag (approved):
Approving with a minor comment

PR #587, opened 2022-09-27
Use truly unreachable address in ConnectionManagerTranslateTest

mdumandag:
As reported by Serkan on a similar issue for Node.js, the `192.168.0.1` is reachable from time to time in the Windows runners in Github Actions.

Being able to reach this address causes some tests to fail. (See the recent nightly runs for the failing tests)

As a solution, we are using a truly unreachable address, similar to what we did for Node.js client.


mdumandag:
See this run for example https://github.com/hazelcast/hazelcast-python-client/actions/runs/3132622714/jobs/5085152946

yuce (approved):

PR #592, opened 2022-11-03
[API-1014] Correct the FieldType values for Portable.

mehmettokgoz:
The FieldType values for new decimal, time, date, and timezone types and their array corresponding were wrong and this PR fixed them.


mdumandag (approved):
Approving this, but we would probably need to enhance our binary compatibility test for these new types so that the errors could be caught earlier. Can you create a Jira task for it? Or, we can do it in this PR if you want

yuce (approved):

mehmettokgoz:
FYI @mdumandag, I've opened a PR on the Java side for Portable object update and this PR now includes the changes for Python client for the binary compatibility test. 

Java PR: https://github.com/hazelcast/hazelcast/pull/22716

tests/unit/serialization/binary_compatibility/reference_objects.py
```diff
@@ -654,6 +704,37 @@ def __eq__(self, other):
     ),
     "BigInteger": 1314432323232411,
     "BigDecimal": decimal.Decimal("31231.12331"),
+    "BigDecimal[]": [
```
mdumandag, line 707: I think we shouldn't define them here. These are not part of the reference objects (which is the equivalent of `allTestObjects` in Java) We can define them outside, just like we did for things like `_data ` below

tests/unit/serialization/binary_compatibility/reference_objects.py
```diff
@@ -816,6 +907,11 @@ def __eq__(self, other):
     "int[]",
     "long[]",
     "String[]",
+    "BigDecimal[]",
+    "LocalDate[]",
+    "LocalTime[]",
+    "LocalDateTime[]",
+    "OffsetDateTime[]",
```
mdumandag, line 914: Once you define them outside, I think you don't need this anymore

tests/unit/serialization/binary_compatibility/reference_objects.py
```diff
@@ -826,8 +922,18 @@ def skip_on_serialize(object_type):
...
+}
+
 _SKIP_ON_DESERIALIZE_PATTERN = re.compile(r"^.*(Predicate|Aggregator|Projection)$")
 
 
 def skip_on_deserialize(object_type):
-    return _SKIP_ON_DESERIALIZE_PATTERN.match(object_type) is not None
+    return (_SKIP_ON_DESERIALIZE_PATTERN.match(object_type) is not None) or (
+        object_type in _SKIP_ON_DESERIALIZE
+    )
```
mdumandag, line 939: Same as above

tests/unit/serialization/binary_compatibility/reference_objects.py
mehmettokgoz: Fixed at https://github.com/hazelcast/hazelcast-python-client/pull/592/commits/a5ac0727172f5527fae403c60cb2fab2426c360a . Thanks for the clarification, I was desperately excluding them manually 😄

tests/unit/serialization/binary_compatibility/reference_objects.py
mehmettokgoz: Removed at https://github.com/hazelcast/hazelcast-python-client/pull/592/commits/a5ac0727172f5527fae403c60cb2fab2426c360a

mdumandag (approved):

yuce (approved):
Looks good!

PR #597, opened 2022-11-29
Set the Ubuntu version used in workflows to 20.04 [API-1745]

### LINKED ISSUE #596: Python 3.6 on ubuntu nightly test cannot start. [API-1745]
Nightly Ubuntu test for Python 3.6 cannot start with the message:
```
Version 3.6 was not found in the local cache
Error: Version 3.6 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```
Using a prior version of Ubuntu instead of `ubuntu-latest` in the workflow may fix this issue.

mdumandag:
There seems to be a temporary issue on the availability of the Python 3.6 version in the ubuntu-22.04 (latest).

Till that issue is resolved, we should pin the Ubuntu version to 20.04, as it is the latest version that has Python 3.6 today.

closes #596

### DIFF
--- a/.github/workflows/coverage_runner.yml
+++ b/.github/workflows/coverage_runner.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       matrix:
         python-version: [ '3.6', '3.10' ]
-        os: [ ubuntu-latest, windows-latest ]
+        os: [ ubuntu-20.04, windows-latest ]
       fail-fast: false
       
     steps:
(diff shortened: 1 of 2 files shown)


yuce (approved):

mdumandag:
@yuce can you force merge this, please? because we use the `pull_request_target`, the changes to the workflows are not picked up in the runners. also, after merging this we might need to update the merge checks in Github because the workflow name will be changed from `Run tests with Python 3.6 on ubuntu-latest` to `Run tests with Python 3.6 on ubuntu-20.04`, I suppose (same for the 3.10 version check)

yuce:
@mdumandag Sure, merged. Thanks for the PR.

Issue #591, opened 2022-11-02
Custom serializer lookup is broken [API-1669]


yuce:
Reproducer:
```
class Foo:
    pass

client = HazelcastClient()
m = client.get_map("test").blocking()
m.set("foo", Foo)
```

Fails with:
```
File "/home/yuce/Work/Projects/hazelcast-python-client/hazelcast/serialization/service.py", line 447, in lookup_custom_serializer
    for super_type in obj_type.__subclasses__():
hazelcast.errors.HazelcastSerializationError: unbound method type.__subclasses__() needs an argument
```

PR #603, opened 2022-12-29
Fix custom lookup serializer for class error [API-1669]

fatihozer0:
closes https://github.com/hazelcast/hazelcast-python-client/issues/591

### DIFF
--- a/hazelcast/serialization/service.py
+++ b/hazelcast/serialization/service.py
@@ -389,9 +389,6 @@ class SerializerRegistry:
         # primitives, arrays, String, UUID and some helper types(BigInteger etc)
         serializer = self.lookup_default_serializer(obj_type, obj)
 
-        if serializer is None and obj_type is type:
-            serializer = self._python_serializer
-
         # 3-Custom registered types by user
         if serializer is None:
             serializer = self.lookup_custom_serializer(obj_type)
@@ -447,7 +444,8 @@ class SerializerRegistry:
         serializer = self._type_dict.get(obj_type, None)
         if serializer is not None:
             return serializer
-        for super_type in obj_type.__subclasses__():
+
+        for super_type in obj_type.__mro__:
             serializer = self.register_from_super_type(obj_type, super_type)
             if serializer is not None:
                 return serializer
--- a/hazelcast/serialization/service.py
+++ b/hazelcast/serialization/service.py
@@ -389,6 +389,9 @@ class SerializerRegistry:
         # primitives, arrays, String, UUID and some helper types(BigInteger etc)
         serializer = self.lookup_default_serializer(obj_type, obj)
 
+        if serializer is None and obj_type is type:
+            serializer = self._python_serializer
+
         # 3-Custom registered types by user
         if serializer is None:
             serializer = self.lookup_custom_serializer(obj_type)
--- a/hazelcast/serialization/service.py
+++ b/hazelcast/serialization/service.py
@@ -6,7 +6,7 @@ import uuid
 
 import typing
 
-from hazelcast.config import IntType, _Config
+from hazelcast.config import IntType, Config
 from hazelcast.errors import HazelcastInstanceNotActiveError
 from hazelcast.serialization.api import IdentifiedDataSerializable, Portable
 from hazelcast.serialization.compact import (
@@ -320,7 +320,7 @@ class SerializationServiceV1:
 class SerializerRegistry:
     def __init__(
         self,
-        config: _Config,
+        config: Config,
         portable_serializer: PortableSerializer,
         data_serializer: IdentifiedDataSerializer,
         compact_serializer: CompactStreamSerializer,
--- a/tests/unit/serialization/custom_global_serialization_test.py
+++ b/tests/unit/serialization/custom_global_serialization_test.py
@@ -42,12 +42,8 @@ class CustomClass:
         return False
 
 
-class TheOtherCustomClass(CustomClass):
-    def __init__(self, uid, name, text, source=None):
-        super(TheOtherCustomClass, self).__init__(uid, name, text, source)
-
-    def __eq__(self, other):
-        return super(TheOtherCustomClass, self).__eq__(other)
+class ChildCustomClass(CustomClass):
+    pass
 
 
 class CustomSerializer(StreamSerializer):
@@ -142,24 +138,19 @@ class CustomSerializationTestCase(unittest.TestCase):
         config.custom_serializers = {CustomClass: CustomSerializer}
         service = SerializationServiceV1(config)
 
-        data1 = service.to_data(CustomClass)
-        deserialized1 = service.to_object(data1)
-
-        self.assertEqual(CustomClass, deserialized1)
-
-        data2 = service.to_data(TheOtherCustomClass)
-        deserialized2 = service.to_object(data2)
-
-        self.assertEqual(TheOtherCustomClass, deserialized2)
+        for clazz in (CustomClass, ChildCustomClass):
+            data = service.to_data(clazz)
+            deserialized = service.to_object(data)
+            self.assertEqual(clazz, deserialized)
 
     def test_serializing_child_class_instances_with_super_class_serializer(self):
         config = Config()
-        config.custom_serializers = {TheOtherCustomClass: TheOtherCustomSerializer}
+        config.custom_serializers = {CustomClass: CustomSerializer}
         service = SerializationServiceV1(config)
 
-        obj = TheOtherCustomClass("uid", "some name", "description text", "CUSTOM")
+        obj = ChildCustomClass("uid", "some name", "description text", "CUSTOM")
         data = service.to_data(obj)
         deserialized = service.to_object(data)
 
-        self.assertTrue(isinstance(deserialized, CustomClass))
+        self.assertIsInstance(deserialized, CustomClass)
         self.assertEqual(obj, deserialized)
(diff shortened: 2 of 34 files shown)


hazelcast/serialization/service.py
```diff
@@ -389,6 +389,9 @@ def serializer_for(self, obj):
         # primitives, arrays, String, UUID and some helper types(BigInteger etc)
         serializer = self.lookup_default_serializer(obj_type, obj)
 
+        if serializer is None and obj_type is type:
```
mdumandag, line 392: We should do it not here but in the broken place. (in `lookup_custom_serializer`)

BTW, the actual problem in `lookup_custom_serializer` is that it considers subclasses whereas it should check the superclasses. If you look at the Java implementation you will see that we consider super types. I would suggest looking there and fixing the actual problem. Then, this will be fixed as well.

fatihozer0: I see. Done now.

hazelcast/serialization/service.py
```diff
@@ -444,7 +444,8 @@ def lookup_custom_serializer(self, obj_type):
         serializer = self._type_dict.get(obj_type, None)
         if serializer is not None:
             return serializer
-        for super_type in obj_type.__subclasses__():
+
+        for super_type in obj_type.__mro__:
```
mdumandag, line 448: Can you add tests for this fix? You can add them to `CustomSerializationTestCase`

fatihozer0: Okay. Working on it.

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +139,21 @@ def test_double_register_custom_serializer(self):
 
         with self.assertRaises(ValueError):
             service._registry.safe_register_serializer(TheOtherCustomSerializer, CustomClass)
+
+
+    def test_class(self):
```
mdumandag, line 144:
```suggestion
    def test_serializing_class_instances(self):
```

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +139,21 @@ def test_double_register_custom_serializer(self):
 
         with self.assertRaises(ValueError):
             service._registry.safe_register_serializer(TheOtherCustomSerializer, CustomClass)
+
```
mdumandag, line 142: unnecessary newline. also, there are some unnecessary new lines in the test method below (lines 145 and 147)

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +139,21 @@ def test_double_register_custom_serializer(self):
...
+
+    def test_class(self):
+
+        config = _Config()
+
+        config.custom_serializers = {CustomClass: CustomSerializer, TheOtherCustomClass: CustomSerializer}
```
mdumandag, line 148: Only one of them would suffice here

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +139,21 @@ def test_double_register_custom_serializer(self):
...
+        service = SerializationServiceV1(config)
+
+        data1 = service.to_data(CustomClass)
+        deserialized1 = service.to_object(data1)
+
+        data2 = service.to_data(TheOtherCustomClass)
+        deserialized2 = service.to_object(data2)
+
+        self.assertEqual(CustomClass, deserialized1)
+        self.assertEqual(TheOtherCustomClass, deserialized2)
```
mdumandag, line 159: We need to add a test where we add a serializer for just `CustomClass`, and test serializing instances of `TheOtherCustomClass` (not the class, but actual objects of type `TheOtherCustomClass`). So, we can be sure that the change we did works fine

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +139,21 @@ def test_double_register_custom_serializer(self):
 
         with self.assertRaises(ValueError):
             service._registry.safe_register_serializer(TheOtherCustomSerializer, CustomClass)
+
+
+    def test_class(self):
+
+        config = _Config()
```
mdumandag, line 146: Please rebase your branch, there is no `_Config` anymore, it is renamed to `Config` and made public

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: Okay, fixed them.

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: Why? They have same superclass but they are different. That is what I thought.

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: Done.

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: It looks like there is a problem. 
```
Expected :<tests.unit.serialization.custom_global_serialization_test.TheOtherCustomClass object at 0x1110d4b20>
Actual   :<tests.unit.serialization.custom_global_serialization_test.CustomClass object at 0x1110d4af0>
```

mdumandag: That looks fine to me, because we have used the serializer of a superclass, so it is expected to receive instances of a superclass in this case. Not the most intuitive design, but at least it is the same as Java

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +140,34 @@ def test_double_register_custom_serializer(self):
...
+
+    def test_serializing_class_and_object_instances(self):
+        config = Config()
+        config.custom_serializers = {CustomClass: CustomSerializer}
+        service = SerializationServiceV1(config)
+
+        data1 = service.to_data(CustomClass)
+        deserialized1 = service.to_object(data1)
+
+        self.assertEqual(CustomClass, deserialized1)
```
mdumandag, line 167: you already tested that in the above. so, we can remove these three lines

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +140,34 @@ def test_double_register_custom_serializer(self):
...
+
+        data2 = service.to_data(TheOtherCustomClass)
+        deserialized2 = service.to_object(data2)
+
+        self.assertEqual(TheOtherCustomClass, deserialized2)
+
+    def test_serializing_class_and_object_instances(self):
```
mdumandag, line 159:
```suggestion
    def test_serializing_child_class_instances_with_super_class_serializer(self):
```

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -42,6 +42,18 @@ def __eq__(self, other):
         return False
 
 
+class TheOtherCustomClass(CustomClass):
+
```
mdumandag, line 46: unnecessary new line

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -42,6 +42,18 @@ def __eq__(self, other):
...
+class TheOtherCustomClass(CustomClass):
+    
+    def __init__(self, uid, name, text, source=None):
+        super(TheOtherCustomClass, self).__init__(uid, name, text, source)
+
+    def __eq__(self, other):
+        super(TheOtherCustomClass, self).__eq__(other)
+
+    def func(self):
+        pass
```
mdumandag, line 54: do we need this func?

mdumandag (commented):
@fatihozer0 can you also take a look at the linter problems?

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: Done.

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: No but wanted to add something to show it is different in some way. Gonna delete.

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: How do we determine the test names?

tests/unit/serialization/custom_global_serialization_test.py
fatihozer0: I just added to show if there is difference in one test. Removing them

tests/unit/serialization/custom_global_serialization_test.py
mdumandag: Well, we try to be as precise as we can so that we can know what the test does in the future by reading the test name. The former name didn't have the necessary info IMO

mdumandag:
@fatihozer0 after you get the test failure fixed, we are ready to merge this

tests/unit/serialization/custom_global_serialization_test.py
```diff
@@ -128,3 +136,30 @@ def test_double_register_custom_serializer(self):
...
+        self.assertEqual(CustomClass, deserialized1)
+
+        data2 = service.to_data(TheOtherCustomClass)
+        deserialized2 = service.to_object(data2)
+
+        self.assertEqual(TheOtherCustomClass, deserialized2)
+
+    def test_serializing_child_class_instances_with_super_class_serializer(self):
+        config = Config()
+        config.custom_serializers = {TheOtherCustomClass: TheOtherCustomSerializer}
```
mdumandag, line 157: We should use the `{CustomClass: CustomSerializer}` here, because in the test name, we are saying that we are going to serialize child class instance with super class serializer.

PR #608, opened 2023-02-23
Improve test Infrastructure [API-1510]

fatihozer0:
PR closed by Hazelcast automation as no activity (>3 months). Please reopen with comments, if necessary. Thank you for using Hazelcast and your valuable contributions

### DIFF
--- a/hazelcast/config.py
+++ b/hazelcast/config.py
@@ -257,7 +257,6 @@ class TopicOverloadPolicy:
 
 
 _DEFAULT_CLUSTER_NAME = "dev"
-_DEFAULT_CLUSTER_PORT = 0
 _DEFAULT_CONNECTION_TIMEOUT = 5.0
 _DEFAULT_RETRY_INITIAL_BACKOFF = 1.0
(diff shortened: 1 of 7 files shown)


yuce:
@fatihozer0 Could you fix the linter issues?

hazelcast/config.py
```diff
@@ -435,6 +438,21 @@ def cluster_name(self, value: str) -> None:
 
         self._cluster_name = value
 
+    @property
```
mdumandag, line 441: Just a note, we should not add a configuration for this task. You should use the port you selected as the port of the addresses in the `cluster_members` config option

fatihozer0: I did not get what do you mean

mdumandag: We use the port from the cluster member addresses and should put that into there. 
```python
client = HazelcastClient(cluster_members=["localhost:12345"])
```
when you do your changes, you need to put the port you selected above, not into a new config you created here.

fatihozer0: Okay, reverting commit then.

PR #609, opened 2023-03-06
Kickoff 5.2.0 development [API-1564]

mdumandag:
This PR contains various fixes to make the client compatible with the 5.2 server.

- Server version used in tests bumped to 5.2.2
- For Compact, field kinds are updated with the up-to-date enum values and they are moved to the serialization.api from compact, to match with the Java side
- Dependency on the enterprise-tests JAR is removed and the self-signed certificates that we depend on moved to the client-side, instead of adding a dependency to the private test artifacts.
- Some reactor related tests that fail on macOS are fixed


hazelcast/serialization/api.py
```diff
@@ -2593,3 +2594,57 @@ def get_type_name(self) -> str:
...
+    ARRAY_OF_TIME = 22
+    DATE = 23
+    ARRAY_OF_DATE = 24
+    TIMESTAMP = 25
+    ARRAY_OF_TIMESTAMP = 26
+    TIMESTAMP_WITH_TIMEZONE = 27
+    ARRAY_OF_TIMESTAMP_WITH_TIMEZONE = 28
+    COMPACT = 29
+    ARRAY_OF_COMPACT = 30
+    NULLABLE_BOOLEAN = 33
```
yuce, line 2637: Adding unsupported fields and 31, 32 may be nice as documentation.

tests/unit/reactor_test.py
```diff
@@ -182,24 +183,29 @@ def test_wake(self):
         self.assertFalse(waker.awake)
         waker.wake()
         self.assertTrue(waker.awake)
+        # Wait until the reader becomes readable
+        select.select([waker._reader], [], [], 10)
```
yuce, line 187: Would it make sense to move this to its own function, like: `waker.wait_reader()`?

yuce (commented):
Looks good, just two nits.

tests/unit/reactor_test.py
mdumandag: I need this just for tests, I don't want to add a function for such a single liner

hazelcast/serialization/api.py
mdumandag: OK, let me do it. I will also do the same for char and array of char

yuce (dismissed):

mdumandag:
@yuce can you re-approve, please?

yuce (approved):

PR #610, opened 2023-03-07
Introduce various updates for the Compact serialization [API-1564]

mdumandag:
This PR introduces couple of updates to the implementation

- Schema now holds a regular dict, instead of OrderedDict for faster lookups and simpler implementation
- Prevent duplicate field names by adding a check to the schema writer
- Disallowing compact serializers to override builtin serializer

closes https://github.com/hazelcast/hazelcast-python-client/issues/559
closes https://github.com/hazelcast/hazelcast-python-client/issues/579

### DIFF
--- a/hazelcast/serialization/compact.py
+++ b/hazelcast/serialization/compact.py
@@ -1,5 +1,4 @@
 import abc
-import collections
 import datetime
 import decimal
 import typing
@@ -1561,6 +1560,7 @@ class SchemaWriter(CompactWriter):
     def __init__(self, type_name: str):
         self._type_name = type_name
         self._fields: typing.List[FieldDescriptor] = []
+        self._field_names: typing.Set[str] = set()
 
     def build(self) -> "Schema":
         return Schema(self._type_name, self._fields)
@@ -1740,16 +1740,21 @@ class SchemaWriter(CompactWriter):
         self._add_field(field_name, FieldKind.ARRAY_OF_COMPACT)
 
     def _add_field(self, name: str, kind: "FieldKind"):
+        if name in self._field_names:
+            raise HazelcastSerializationError(f"Field with the name '{name}' already exists")
+
+        self._field_names.add(name)
         self._fields.append(FieldDescriptor(name, kind))
 
 
 class Schema:
     def __init__(self, type_name: str, fields_list: typing.List["FieldDescriptor"]):
         self.type_name = type_name
-        self.fields: typing.Dict[str, "FieldDescriptor"] = Schema._dict_to_key_ordered_dict(
-            {f.name: f for f in fields_list}
-        )
-        self.fields_list = list(self.fields.values())
+        self.fields: typing.Dict[str, "FieldDescriptor"] = {f.name: f for f in fields_list}
+        # Sort the fields by the field name so that the field offsets/indexes
+        # can be set correctly.
+        fields_list.sort(key=lambda f: f.name)
+        self.fields_list = fields_list
         self.schema_id: int = 0
         self.var_sized_field_count: int = 0
         self.fix_sized_fields_length: int = 0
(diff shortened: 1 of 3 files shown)


hazelcast/serialization/compact.py
```diff
@@ -1769,6 +1774,11 @@ def _init(self):
             else:
                 fix_sized_fields.append(field)
 
+        # Fix sized fields should be in descending order of size in bytes.
```
yuce, line 1777: Why are they sorted by size in bytes?

hazelcast/serialization/service.py
```diff
@@ -496,6 +501,22 @@ def register_from_super_type(self, obj_type, super_type) -> typing.Optional[Stre
...
+        """
+        Makes sure that the classes registered as Compact serializable are not
+        overriding the default serializers.
+
+        Must be called in the constructor of the serialization service after it
+        completes registering default serializers.
+        """
+        for compact_type in self._compact_types:
+            if compact_type in self._constant_type_dict:
+                raise HazelcastSerializationError(
```
yuce, line 514: Java client throws `IllegalArgumentException` here.

yuce (commented):
Looks good, just some ideas.

hazelcast/serialization/compact.py
```diff
@@ -1561,6 +1560,7 @@ class SchemaWriter(CompactWriter):
     def __init__(self, type_name: str):
         self._type_name = type_name
         self._fields: typing.List[FieldDescriptor] = []
+        self._field_names: typing.Set[str] = set()
```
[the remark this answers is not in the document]
mdumandag: The order is not important here, fields are sorted in the Schema. And, it expects a list, not an OrderedDict. I could make use of a single map here(from field name to field descriptor), and pass list(map.values()) to the Schema which is essentially the same thing with this

hazelcast/serialization/compact.py
mdumandag: That was a requirement of the wire-level format https://hazelcast.atlassian.net/wiki/spaces/IMDG/pages/3507487040/Compact+Wire-level+Format+Design+Document#Fixed-Size-Fields

hazelcast/serialization/service.py
mdumandag: That makes more sense, let me change it 👍 BTW, do you think using `ValueError` in Python for this case, instead of `IllegalArgumentError` (which is one of our custom errors) would be more idiomatic?

hazelcast/serialization/compact.py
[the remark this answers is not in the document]
yuce: I think the solution with the dict and passing `map.values()` would be a better solution, you can also pass it directly to `Schema` to remove the `self.fields: typing.Dict[str, "FieldDescriptor"] = {f.name: f for f in fields_list}` line.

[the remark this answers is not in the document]
mdumandag: OK, I will make use of a single dict here, but won't change the schema init function, as this is not the only place I pass something to it. The protocol passes a list to it (see SchemaCodec)

hazelcast/serialization/service.py
yuce: I'm leaning towards `IllegalArgumentError` to make it more similar to the reference implementation. In the Python docs there's this at: https://docs.python.org/3/library/exceptions.html
```
Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).
```
So, `ValueError` seems to be more appropriate when there's a problem with the value itself.

hazelcast/serialization/compact.py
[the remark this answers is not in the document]
mdumandag: https://github.com/hazelcast/hazelcast-python-client/pull/610/commits/b4d0105c9b4f075e904e92841fd3cc8ca39526f6

hazelcast/serialization/service.py
mdumandag: https://github.com/hazelcast/hazelcast-python-client/pull/610/commits/b4d0105c9b4f075e904e92841fd3cc8ca39526f6

yuce (approved):
Looks good!

PR #611, opened 2023-03-08
Update Compact documentation [API-1564]

mdumandag:
This PR updates the Compact related documentation.

- Corrects a few stale places in the API docs
- Removes BETA annotations from the API docs
- Adds a documentation about the nullable-primitive interoperability

This PR does not remove the limitations section in the reference docs so that we can use the list there to make sure that the Compact works nicely with those APIs, before releasing 5.2.

closes https://github.com/hazelcast/hazelcast-python-client/issues/565
closes https://github.com/hazelcast/hazelcast-python-client/issues/582

### DIFF
--- a/docs/serialization.rst
+++ b/docs/serialization.rst
@@ -264,6 +264,9 @@ and types. Hazelcast uses the write method of the serializer to extract a
 schema out of the object, hence any conditional code that may or may not run
 depending on the object in that method might result in an undefined behavior.
 
+Additionally, evolved serializers must have the same type name with the
+initial version of the serializer.
+
 Limitations
 ~~~~~~~~~~~
 
--- a/docs/serialization.rst
+++ b/docs/serialization.rst
@@ -82,13 +82,8 @@ use a :ref:`serialization:custom serialization`.
 Compact Serialization
 ---------------------
 
-.. warning::
-    Compact Serialization feature is in the BETA status and any part of it
-    might be changed without a prior notice, until it is promoted to the
-    stable status.
-
-As an enhancement to existing serialization methods, Hazelcast offers a beta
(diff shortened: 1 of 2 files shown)


docs/serialization.rst
```diff
@@ -260,6 +257,13 @@ Note that, when an old reader reads data written by an old writer, or a new
 reader reads a data written by a new writer, they will be able to read all
 fields.
 
+One thing to be careful while evolving the class is to not have any conditional
```
yuce, line 260: In the `get_type_name` 's documentation there's some info about using the same type name between evolved types, should we also mention that here?

hazelcast/serialization/api.py
```diff
@@ -1366,36 +1366,54 @@ def get_raw_data_output(self) -> ObjectDataOutput:
...
         Returns:
-              Field kind for the given field. or FieldKind.NOT_AVAILABLE if the field does not exist.
+            Kind of the field.
         """
 
     @abc.abstractmethod
     def read_boolean(self, field_name: str) -> bool:
         """Reads a boolean.
 
+        This method can also read a nullable boolean, as long as it is not
```
yuce, line 1414: I am not sure about encouraging the user to use `read_boolean` instead of `read_nullable_boolean`. Would it be better to say something like: "Prefer to use read_nullable_boolean if it is possible that the read value is None"

yuce (commented):
Looks good, just some ideas...

docs/serialization.rst
mdumandag: https://github.com/hazelcast/hazelcast-python-client/pull/611/commits/66d9811255e30a1abc955f72957793e2b9cbedb0

yuce (approved):

PR #617, opened 2023-03-15
Do not invoke urgent invocations that contain serialized data immediately [API-1564]

mdumandag:
Although we believe it was a mistake, the urgent invocations on the
client-side are used for some user invocations like listener registrations,
apart from the actual urgent invocations like heartbeats, authentication, etc.

When the client reconnects to some cluster, it sends the local state
in some executor. So, there might be some time between the local state is
sent. However, during that time, urgent invocations are allowed to
go through from the client. That might violate our assumption that
the schema is received by the cluster before the data.

To solve this, we will not invoke urgent invocations that contain
serialized data if the client is not initialized on the cluster, and there
were some compact schemas sent to the previous clusters. Such
invocations are only related to user invocations so we will not delay
invoking actual urgent invocations like heartbeats.

protocol pr: https://github.com/hazelcast/hazelcast-client-protocol/pull/459
closes https://github.com/hazelcast/hazelcast-python-client/issues/578


yuce (approved):
Looks good!

PR #618, opened 2023-03-20
Eagerly deserialize responses consisting of list of data [API-1964]

mdumandag:
We were performing lazy deserialization on APIs that return a list of data, such as map#values(). This was not going to work with Compact, as after returning the list to the user, there might be Compact serialized data on the list which we don't have the schema on the client yet. If it was a normal response, the client would have fetched the schema, but it is not possible with these lazy APIs. We perform the deserialization while getting the list items and this is a sync API. We can't perform a blocking call there, because there is a chance that this will be executed in the reactor thread, which would result in a deadlock.

So, the only possible way of making these APIs work with Compact is to convert them to eager deserialization.

This PR removes lazy deserialization in everything apart from SQL, which will be handled in another PR.

### DIFF
--- a/hazelcast/util.py
+++ b/hazelcast/util.py
@@ -1,12 +1,13 @@
 import random
 import threading
 import time
+import typing
 import uuid
-
-from collections.abc import Sequence, Iterable
 
 from hazelcast.serialization import UUID_MSB_SHIFT, UUID_LSB_MASK, UUID_MSB_MASK
 
+if typing.TYPE_CHECKING:
+    from hazelcast.serialization.data import Data
 
 DEFAULT_ADDRESS = "127.0.0.1"
 DEFAULT_PORT = 5701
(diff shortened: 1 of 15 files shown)


hazelcast/util.py
```diff
@@ -175,6 +135,21 @@ def get_portable_version(portable, default_version):
     return version
 
 
+def deserialize_list_in_place(data_list, to_object_fn):
```
yuce, line 138: Should there be type annotation for the new functions?

yuce (commented):
Looks good, just a minor idea...

hazelcast/util.py
mdumandag: That makes sense, I wish we fully type hint the code base so that we can add checks for such things.

mdumandag: https://github.com/hazelcast/hazelcast-python-client/pull/618/commits/c8ff71bc35d74b6655c8125ac7d814a00c79814c

yuce (approved):

PR #619, opened 2023-03-20
Eagerly deserialize SQL columns of type OBJECT [API-1964]

mdumandag:
Due to lazy deserialization, Compact serialized objects were
not working nicely with SQL. It was possible to get deserialization
errors that cannot be easily understood by the user.

Similar to what we did for other APIs, we have removed lazy
deserialization from this API as well.

To do that efficiently, I have modified the protocol code generator
so that we can pass to_object function to the SqlPageCodec where
we can do deserialization while filling rows and columns, so that
we will not iterate over the result set once more.

protocol pr: https://github.com/hazelcast/hazelcast-client-protocol/pull/461

### DIFF
--- a/hazelcast/protocol/builtin.py
+++ b/hazelcast/protocol/builtin.py
@@ -1,3 +1,4 @@
+import typing
 import uuid
 from datetime import date, time, datetime, timedelta, timezone
 from decimal import Decimal
@@ -701,7 +702,7 @@ class BigDecimalCodec:
 
 class SqlPageCodec:
     @staticmethod
-    def decode(msg):
+    def decode(msg, to_object_fn):
         from hazelcast.sql import SqlColumnType, _SqlPage
 
         # begin frame
@@ -715,7 +716,7 @@ class SqlPageCodec:
         column_count = len(column_type_ids)
 
         # read columns
-        columns = [None] * column_count
+        columns: typing.List = [None] * column_count
 
         for i in range(column_count):
             column_type_id = column_type_ids[i]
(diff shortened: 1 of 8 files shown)


hazelcast/protocol/builtin.py
```diff
@@ -701,7 +702,7 @@ def decode(msg):
 
 class SqlPageCodec:
     @staticmethod
-    def decode(msg):
+    def decode(msg, to_object_fn):
         from hazelcast.sql import SqlColumnType, _SqlPage
```
yuce, line 706: Is this import here in order to prevent circular imports?
If not, it would be better to move it to global in order to not to pay the (very small) cost.

hazelcast/serialization/portable/serializer.py
```diff
@@ -53,18 +53,18 @@ def find_portable_version(self, factory_id, class_id, portable):
...
+        portable_factory = self._portable_factories.get(factory_id)
+        if not portable_factory:
             raise HazelcastSerializationError(
                 "Could not find portable_factory for factory-id: %s" % factory_id
             )
 
-        portable = portable_factory[class_id]
-        if portable is None:
+        portable = portable_factory.get(class_id)
+        if not portable:
```
yuce, line 63: Checking with `is None` looks more correct to me.

yuce (commented):
Looks good, two minor comments.

hazelcast/protocol/builtin.py
mdumandag: Yes, it was due to circular imports

hazelcast/serialization/portable/serializer.py
mdumandag: https://github.com/hazelcast/hazelcast-python-client/pull/619/commits/83eeea0e7068e38f06f360531cb84d0108f4a5bb

yuce (approved):

PR #637, opened 2023-06-20
Update Let's Encrypt certificate periodically

mdumandag:
This PR adds a workflow to update the Let's Encrypt certificate used in tests periodically.

The workflow updates the certificate used in tests every month, based on the certificate generated in the private test artifacts repo. The certificate is renewed every week there, and each certificate is valid for 3 months, so renewing the certificate in this repo every month should work without a problem.

### DIFF
--- a/.github/workflows/update_lets_encrypt_certificate.yml
+++ b/.github/workflows/update_lets_encrypt_certificate.yml
@@ -1,0 +1,39 @@
+name: Update Let's Encrypt certificate
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


mdumandag:
For reference, this is a commit generated by this workflow:

[b0b6fd7](https://github.com/hazelcast/hazelcast-python-client/commit/b0b6fd7b14347fdd9ecb98e51aa8e2d5da101e3c)

srknzl (approved):

PR #642, opened 2023-08-07
Update dependency:get command to latest syntax

JackPGreen:
The `python run_tests.py` script cannot be run as it has the same problem as described in https://github.com/hazelcast/hazelcast-cpp-client/pull/1206

### DIFF
--- a/start_rc.py
+++ b/start_rc.py
@@ -48,7 +48,7 @@ def download_if_necessary(repo, artifact_id, version, is_test_artifact=False):
         "-q",
         "dependency:get",
         "-Dtransitive=false",
-        "-DrepoUrl=" + repo,
+        "-DremoteRepositories=" + repo,
         "-Dartifact=" + artifact,
         "-Ddest=" + dest_file_name,
     ]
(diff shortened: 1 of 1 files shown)


mehmettokgoz (changes requested):
It looks like `-Ddest` is removed after Maven 3.x and script does not downloads JARs inside repo folder so RC fails to start.

start_rc.py
```diff
@@ -46,9 +46,10 @@ def download_if_necessary(repo, artifact_id, version, is_test_artifact=False):
     args = [
         "mvn",
         "-q",
-        "dependency:get",
+        "org.apache.maven.plugins:maven-dependency-plugin:2.10:get",
         "-Dtransitive=false",
-        "-DrepoUrl=" + repo,
+        "-DremoteRepositories=" + repo,
+        "-Dartifact=" + artifact,
```
mehmettokgoz, line 52: This line is duplicated.
```suggestion
```

mehmettokgoz (changes requested):

ihsandemir (approved):

mehmettokgoz (approved):

Issue #658, opened 2024-01-15
`linter_docs_mypy.yaml` action fails due to `Sphinx version error`


JackPGreen:
[My PR's](https://github.com/hazelcast/hazelcast-python-client/pull/657) linter [status check has failed](https://github.com/hazelcast/hazelcast-python-client/actions/runs/7531939745/job/20501588880?pr=657) with the following message:
```
sphinx-build -b html -d _build/doctrees  -W --keep-going -b linkcheck . _build/html
Running Sphinx v4.2.0

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
make: *** [Makefile:5[8](https://github.com/hazelcast/hazelcast-python-client/actions/runs/7531939745/job/20501588880?pr=657#step:5:9): html] Error 2
Error: Process completed with exit code 2.
```

PR #659, opened 2024-01-15
Upgrade Sphinx

JackPGreen:
Fixes https://github.com/hazelcast/hazelcast-python-client/issues/658

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 -r requirements-test.txt
 black==22.3.0
-Sphinx==4.2.0
+Sphinx==7.2.6
 sphinx-rtd-theme==1.0.0
 mypy==0.940


docs/conf.py
```diff
@@ -84,7 +84,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
```
srknzl, line 87: not sure what this does?

srknzl (approved):
I built html, it seems fine

docs/conf.py
JackPGreen: The new version of Sphix does not allow `None`, but previously `None` evaluated to `en` anyway.

https://github.com/sphinx-doc/sphinx/issues/10474

PR #664, opened 2024-02-09
Fix compatibility tests 

srknzl:
* Skips paging predicate tests and skips on_cancel count check in older than 5.4

### DIFF
--- a/tests/integration/backward_compatible/predicate_test.py
+++ b/tests/integration/backward_compatible/predicate_test.py
@@ -28,7 +28,7 @@ from tests.integration.backward_compatible.util import (
     write_string_to_writer,
     read_string_from_reader,
 )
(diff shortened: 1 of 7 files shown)


tests/integration/backward_compatible/predicate_test.py
```diff
@@ -171,6 +171,8 @@ def test_false(self):
         self.assertCountEqual(self.map.key_set(predicate), [])
 
     def test_paging(self):
+        # https://hazelcast.atlassian.net/browse/API-2187
```
ihsandemir, line 174: Os repo, better if we have issue number in github for community.

ihsandemir (dismissed):

ihsandemir (approved):

PR #674, opened 2024-05-03
Initial vector search implementation

yuce:


start_rc.py
```diff
@@ -64,18 +65,28 @@ def download_if_necessary(repo, artifact_id, version, is_test_artifact=False):
...
+        server = download_if_necessary(
+            ENTERPRISE_REPO, HAZELCAST_GROUP, "hazelcast-enterprise", SERVER_VERSION
+        )
+        vector = download_if_necessary(
+            ENTERPRISE_REPO, HAZELCAST_GROUP, "hazelcast-enterprise-vector", SERVER_VERSION
+        )
+        artifacts.append(vector)
+        jvector = download_if_necessary(RELEASE_REPO, "io.github.jbellis", "jvector", "2.0.5")
+        artifacts.append(jvector)
+        math3 = download_if_necessary(RELEASE_REPO, "org.apache.commons", "commons-math3", "3.6.1")
```
srknzl, line 86: I'd say let's put these versions at the top of the file, next to SERVER_VERSION &
RC_VERSION.

Why do we need these jars though? Should not "hazelcast-enterprise-vector" include them(any dependency needed for vectors) already? If you are going to shade those dependencies into hazelcast-enterprise-vector soon, I'd add a todo comment here about removing these dependencies from here because it seems this pr will be open for a while until 5.4 is released

examples/vector_collection/vector_collection.py
```diff
@@ -0,0 +1,73 @@
+import logging
+
+from hazelcast import vector
+from hazelcast.core import HazelcastJsonValue
```
srknzl, line 4:
```suggestion
```
unused

examples/vector_collection/vector_collection.py
```diff
@@ -0,0 +1,73 @@
...
+    # Connect to the cluster
+    cfg = Config()
+    cfg.cluster_members = ["localhost:5701"]
+    client = HazelcastClient(cfg)
+
+    vc_name = "my-vector-collection"
+
+    # Create the VectorCollection
+    indexes = [
+        vector.IndexConfig(name="default-vector", metric=vector.Metric.COSINE, dimension=2),
```
srknzl, line 23: nit: 

the following can be better in terms of readability & consistency

```suggestion
       IndexConfig(name="default-vector", metric=Metric.COSINE, dimension=2),
```

hazelcast/vector.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+__all__ = "Document", "Vector", "IndexConfig", "SearchResult"
+
+
+class Type(enum.IntEnum):
+    DENSE = 0
+
+
+VectorType = Type
```
srknzl, line 13: I wonder why we need these variables 🤔 can't we just import Type in other modules for example? (as I see VectorType is used for importing)

hazelcast/vector.py
srknzl, line 1: can you please add docstring to all publicly visible (and even better, for internal ones as well) classes here?

examples/vector_collection/vector_collection.py
```diff
@@ -0,0 +1,73 @@
...
+    ]
+    client.create_vector_collection(vc_name, indexes=indexes)
+
+    # Use the VecotorCollection
```
srknzl, line 27: typo

examples/vector_collection/vector_collection.py
```diff
@@ -0,0 +1,73 @@
...
+    cfg.cluster_members = ["localhost:5701"]
+    client = HazelcastClient(cfg)
+
+    vc_name = "my-vector-collection"
+
+    # Create the VectorCollection
+    indexes = [
+        vector.IndexConfig(name="default-vector", metric=vector.Metric.COSINE, dimension=2),
+    ]
+    client.create_vector_collection(vc_name, indexes=indexes)
```
srknzl, line 25: interesting that we need to do create() and then get() as opposed to other data structures, do we need a create() just for vector collection?

ps: I didn't see java API if there is, can you link it if there is?

examples/vector_collection/vector_collection.py
```diff
@@ -0,0 +1,73 @@
...
+            Vector("default-vector", Type.DENSE, [0.5, 0.7]),
+        ],
+    )
+
+    # Add the Document
+    key = "key-2"
+    vc.set(key, doc2)
+
+    # Search for a vector
+    results = vc.search_near_vector(
```
srknzl, line 53: do we have docstring/documentation of search_near_vector and other methods somewhere? I didn't see them.

hazelcast/client.py
```diff
@@ -373,6 +378,21 @@ def get_topic(self, name: str) -> Topic[MessageType]:
         """
         return self._proxy_manager.get_or_create(TOPIC_SERVICE, name)
 
+    def create_vector_collection(self, name: str, indexes: typing.List[IndexConfig]) -> None:
```
srknzl, line 381: this is adding vector collection config, not creating vector collection, so maybe rename the method

hazelcast/protocol/builtin.py
```diff
@@ -535,6 +539,31 @@ def decode(msg):
         return result
 
 
+class FloatArrayCodec:
```
srknzl, line 542: not related to this pr but I realized most of the code is repeating in this builtin codecs, i wonder why we don't reuse code.

srknzl (commented):
main code & test not reviewed, review in progress

start_rc.py
yuce: Currently `hazelcast-enterprise-vector` jar doesn't include those dependencies.
But eventually it will, so we can just delete those lines when that happens.

hazelcast/vector.py
yuce: I'll add the documentation in a separate PR.

examples/vector_collection/vector_collection.py
yuce: Java API is different from the Python API.
There's currently not a decision about which way to go:
https://github.com/hazelcast/hazelcast-mono/blob/064036486c8da3725b2659995598618715f31904/hazelcast-enterprise-vector/src/main/java/com/hazelcast/vector/VectorCollection.java#L101

hazelcast/protocol/builtin.py
yuce: I guess that would be a separate tech debt cleanup task.

hazelcast/client.py
yuce: That makes sense.
Renamed it to `create_vector_collection_config` at:
https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/3c27230ad7d9777be1698db0d6ceb00121844ba8

examples/vector_collection/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/3c27230ad7d9777be1698db0d6ceb00121844ba8

examples/vector_collection/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/3c27230ad7d9777be1698db0d6ceb00121844ba8

examples/vector_collection/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/3c27230ad7d9777be1698db0d6ceb00121844ba8

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+            doc.value = self._to_data(doc.value)
+            try:
+                entry = (self._to_data(key), doc)
+            except SchemaNotReplicatedError as e:
+                return self._send_schema_and_retry(e, self.put_all, map)
+
+            partition_id = partition_service.get_partition_id(entry[0])
+            try:
+                partition_map[partition_id].append(entry)
+            except KeyError:
```
srknzl, line 73: actually this key error is costly, we should use an explicit if statement here for key existence check

the reason is that if exception is triggers, try/except statements are slower than explicit if statements, see https://stackoverflow.com/a/2522013/21388077

The exception will be raised for sure and for each partition id and multiply that with the number of items in the map.items().

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+    def put_if_absent(self, key: Any, document: Document) -> Future[Optional[Document]]:
+        check_not_none(key, "key can't be None")
+        check_not_none(document, "document can't be None")
+        check_not_none(document.value, "document value can't be None")
+        return self._put_if_absent_internal(key, document)
+
+    def search_near_vector(
+        self,
+        vector: Vector,
+        *,
```
srknzl, line 93: just curious, does this * serve any purpose, or unnecessary?

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+        include_value: bool = False,
+        include_vectors: bool = False,
+        limit: Optional[int] = None
+    ) -> Future[List[SearchResult]]:
+        check_not_none(vector, "vector can't be None")
+        if limit is None:
+            limit = 2147483629
```
srknzl, line 100: this is 2 ** 31 - 19

I wonder what's special about this number. consider introducing a constant and explain it via constant's name or docstring

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+        *,
+        include_value: bool = False,
+        include_vectors: bool = False,
+        limit: Optional[int] = None
+    ) -> Future[List[SearchResult]]:
+        def handler(message):
+            results: List[
+                SearchResult
+            ] = vector_collection_search_near_vector_codec.decode_response(message)
+            for result in results:
```
srknzl, line 158: what about setting vector.type here? SearchResult includes vector list

srknzl (commented):
Review of vector.py

tests/integration/backward_compatible/proxy/vector_collection_test.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def setUp(self):
+        name = random_string()
+        self.client.create_vector_collection_config(name, [IndexConfig("vector", Metric.COSINE, 3)])
+        self.vector_collection = self.client.get_vector_collection(name).blocking()
+
+    def tearDown(self):
+        self.vector_collection.destroy()
+
+    def test_set(self):
```
srknzl, line 41: I would like to see a get() and assertion of values in this test

tests/integration/backward_compatible/proxy/vector_collection_test.py
```diff
@@ -0,0 +1,151 @@
...
+        doc = Document("v1", Vector("vector", Type.DENSE, [0.1, 0.2, 0.3]))
+        self.vector_collection.set("k1", doc)
+
+    def test_get(self):
```
srknzl, line 45: this can be renamed to get_set and the test above can be removed (test_set)

tests/integration/backward_compatible/proxy/vector_collection_test.py
```diff
@@ -0,0 +1,151 @@
...
+        doc = self.vector_collection.get("k1")
+        self.assertIsNone(doc)
+        doc = Document("v1", Vector("vector", Type.DENSE, [0.1, 0.2, 0.3]))
+        self.vector_collection.set("k1", doc)
+        got_doc = self.vector_collection.get("k1")
+        self.assert_document_equal(got_doc, doc)
+
+    def test_put(self):
```
srknzl, line 53: can we also test, put() actually puts, so we can add a get() and assertion of doc2 is retrieved at the end

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+        if not map:
+            return ImmediateFuture(None)
+        partition_service = self._context.partition_service
+        partition_map: Dict[str, List[Tuple[Data, Document]]] = {}
+
+        for key, doc in map.items():
+            check_not_none(key, "key can't be None")
+            check_not_none(doc, "value can't be None")
+            doc = copy.copy(doc)
+            doc.value = self._to_data(doc.value)
```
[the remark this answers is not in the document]
yuce: `doc.Value` can be `None`.

srknzl (commented):
review of the test

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+    Document,
+    SearchResult,
+    Vector,
+    VectorPair,
+    VectorType,
+    VectorSearchOptions,
+)
+
+
+class VectorCollection(Proxy["BlockingVectorCollection"]):
```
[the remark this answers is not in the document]
yuce: I'll add the documentation in a separate PR.

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+    def __init__(self, service_name, name, context):
+        super(VectorCollection, self).__init__(service_name, name, context)
+        self._reference_id_generator = context.lock_reference_id_generator
```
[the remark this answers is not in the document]
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/304e76ee18d663a8430d64e3de8b2206f09c6d89

hazelcast/proxy/vector_collection.py
yuce: Exceptions are not slow in Python.
We need to benchmark it before concluding the current way is slower.
But I changed that code to be more Pythonic at:
https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/304e76ee18d663a8430d64e3de8b2206f09c6d89

hazelcast/proxy/vector_collection.py
yuce: It forces the user to pass the rest of the arguments as keyword arguments.

hazelcast/protocol/builtin.py
srknzl: yeah asking because you may know python stuff more: for example maybe reusing code is costly due to method extra invocations etc. 🤷🏻‍♂️

hazelcast/proxy/vector_collection.py
```diff
@@ -0,0 +1,282 @@
...
+)
+from hazelcast.proxy import Proxy
+from hazelcast.serialization.compact import SchemaNotReplicatedError
+from hazelcast.serialization.data import Data
+from hazelcast.util import check_not_none
+from hazelcast.vector import (
+    Document,
+    SearchResult,
+    Vector,
+    VectorPair,
```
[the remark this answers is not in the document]
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/da367f9c611eef16c1aaceddaef097a0cf1dc9d3

hazelcast/proxy/vector_collection.py
yuce: I set the default to `-1` like in the Java client.

tests/integration/backward_compatible/proxy/vector_collection_test.py
yuce: `test_get` checks the retrieved value, so I think it's OK that this test just checks whether `set` runs without errors.

tests/integration/backward_compatible/proxy/vector_collection_test.py
yuce: I'd like to keep it as is.

tests/integration/backward_compatible/proxy/vector_collection_test.py
yuce: Since there are two `put`s with the same key, this is already tested.

tests/integration/backward_compatible/proxy/vector_collection_test.py
srknzl: anyway, redundant tests are not a big deal. fine by me

hazelcast/proxy/vector_collection.py
yuce: The stackoverflow answers may not apply universally.
We would have to benchmark different approaches, which I think is out-of-scope in this case.

hazelcast/proxy/vector_collection.py
yuce: Yes, that's a mistake.
Fixed at: https://github.com/hazelcast/hazelcast-python-client/pull/674/commits/560871086ac734fcd0a044da584a8309604daa98

srknzl (dismissed):

srknzl (approved):

start_rc.py
k-jamroz: https://github.com/hazelcast/hazelcast-python-client/pull/679

PR #677, opened 2024-05-30
Use released Hazelcast `5.4.0` instead of `SNAPSHOT`

JackPGreen:
Now `5.4.0` is released, the released version should be used in preferences to a pre-release `SNAPSHOT`.

### DIFF
--- a/start_rc.py
+++ b/start_rc.py
@@ -3,7 +3,7 @@ import subprocess
 import sys
 from os.path import isfile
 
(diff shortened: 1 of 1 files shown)


srknzl (approved):
what we do usually is setting it to the next snapshot, that is 5.5.0-snapshot but this is also fine. When we're about to release 5.5, we can make it 5.5-snapshot

PR #678, opened 2024-06-10
[HZAI-73] Vector API Docs

yuce:
* Added the docs for Vector API classes.
* Synced the defaults of `IndexConfig` with the server-side.


hazelcast/vector.py
```diff
@@ -65,20 +97,35 @@ class Metric(enum.IntEnum):
...
 
     name: str
     metric: Metric
     dimension: int
-    max_degree: int = 200
-    ef_construction: int = 300
-    use_deduplication: bool = False
+    max_degree: int = 16
+    ef_construction: int = 100
+    use_deduplication: bool = True
```
EugeneAbramchuk, line 117: this flag is missing from the args docs above, let's add it?

EugeneAbramchuk, line 117: this flag is missing from the args docs above, let's add it?

hazelcast/proxy/vector_collection.py
```diff
@@ -26,29 +26,117 @@
...
+    The vector collection can work both async(non-blocking) or sync(blocking).
+    Blocking calls return the value of the call and block the execution until return value is calculated.
+    However, async calls return ``Future`` and do not block execution.
+    Result of the ``Future`` can be used whenever ready.
+    A ``Future``'s result can be obtained with blocking the execution by calling ``future.result()``.
+
+    The configuration of the vector collection must exist before it can be used.
+
+    Example:
+        #
```
EugeneAbramchuk, line 41: Looks like a surviving debug char :)

hazelcast/vector.py
```diff
@@ -15,6 +15,16 @@ class Type(enum.IntEnum):
...
 @dataclasses.dataclass
 class Vector:
+    """Vector represents a named vector of floats.
+
+    Args:
+        name: Name of the vector.
+            To use the default vector index, name must be set
+            to the blank string (``""``)
+        type: Type of the vector.
+            Currently only ``Type.DENSE`` is supported.
```
EugeneAbramchuk, line 25: Do we have a vector types description somewhere? I was unable to find it in the java docs either

hazelcast/proxy/vector_collection.py
```diff
@@ -90,6 +197,22 @@ def search_near_vector(
...
+        The search is performed using distance metric set when
+        creating the vector index.
+
+        Args:
+            vector: The vector to be used as the reference.
+                It must have the same dimension as specified when creating the vector index.
+            include_value: Return value attached to the Document.
+            include_vectors: Return vectors attached to the Document.
+            limit: Limit the maximum number of Documents returned.
+                If not set, the server default is used.
```
k-jamroz, line 211: there is no server default, I removed `-1` default from Python client in https://github.com/hazelcast/hazelcast-python-client/pull/679

hazelcast/vector.py
yuce: `DENSE` is just a regular vector. `SPARSE` vectors would be an optimization which are not supported yet.

hazelcast/proxy/vector_collection.py
yuce: I think we should have a server default.

hazelcast/vector.py
yuce: Added at: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/86856148d6d318eb453ee672625196cf2ad11e8c

hazelcast/proxy/vector_collection.py
yuce: Removed at: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/86856148d6d318eb453ee672625196cf2ad11e8c

hazelcast/proxy/vector_collection.py
yuce: I've set it to 10 by default for now.

hazelcast/proxy/vector_collection.py
```diff
@@ -99,10 +224,41 @@ def search_near_vector(
...
+        Returns:
+            The Document associated with key, or ``None`` if there was
+            no mapping for key.
+        """
         check_not_none(key, "key can't be None")
         return self._remove_internal(key)
 
     def delete(self, key: Any) -> Future[None]:
+        """Removes the mapping for a key from this VectorCollection if it is present
+        (optional operation).
```
k-jamroz, line 249: what does it mean "optional operation" in this context? It is used also for in other proxies

yuce: I took it from the `Map` documentation. I think it means the operation is no-op if the mapping is not present.

yuce: `delete` deletes a mapping, it's not related to the proxy lifetime.
Does `VectorCollection.delete` behave differently than `Map.delete` if the mapping doesn't exist?

k-jamroz: > Does VectorCollection.delete behave differently than Map.delete if the mapping doesn't exist?

no, it is the same . I think I misunderstood before: you meant mapping=entry, I was thinking mapping=IMap.

I am not sure how to concisely say that the operation does nothing if there is nothing to do, but this is not what I understand when I read "optional operation". I think "idempotent" would be more precise.

and why `delete` is "optional" while `remove` is not? because the latter returns a value? (BTW, `remove` is not idempotent, because repeated it will return different result).

yuce: As far as I see from the server-side code, the difference between `remove` and `delete` may be related to map stores. `remove` calls `mapDataStore.load(key)` even if a mapping is not found.

yuce: Btw, I don't think `delete` is idempotent either.
Apart from that, even calls like `get` are not truly idempotent in Hz.

yuce:
@k-jamroz @EugeneAbramchuk Could you guys approve the PR if the points you raised were addressed sufficiently?

hazelcast/proxy/vector_collection.py
k-jamroz: vector collection does not support map stores, so there is no difference between `remove` and `delete` that I would call "optional operation". So I'd remove `optional operation` from vector docs not to introduce confusion.

hazelcast/proxy/vector_collection.py
```diff
@@ -29,29 +29,117 @@
 
 
 class VectorCollection(Proxy["BlockingVectorCollection"]):
+    """VectorCollection contains documents with vectors.
+
+    Concurrent, distributed, observable and searchable vector collection.
+    The vector collection can work both async(non-blocking) or sync(blocking).
```
k-jamroz, line 35:
```suggestion
    The vector collection can work both async (non-blocking) or sync (blocking).
```

hazelcast/proxy/vector_collection.py
```diff
@@ -29,29 +29,117 @@
...
+        does not contain this key.
+
+        Warning:
+            This method returns a clone of original value, modifying the
+            returned value does not change the actual value in the VectorCollection. One
+            should put modified value back to make changes visible to all nodes.
+
+                >>> value = my_vc.get(key)
+                >>> value.update_some_property()
+                >>> my_vc.set(key,value)
```
k-jamroz, line 70:
```suggestion
                >>> my_vc.set(key, value)
```

hazelcast/proxy/vector_collection.py
```diff
@@ -29,29 +29,117 @@
...
+        Warning:
+            This method uses ``__hash__`` and ``__eq__`` methods of binary form
+            of the key, not the actual implementations of ``__hash__`` and
+            ``__eq__`` defined in key's class.
+
+        Args:
+            key: The specified key.
+
+        Returns:
+            The value for the specified key.
```
k-jamroz, line 81:
```suggestion
            Document associated with key or ``None`` if there is no mapping
            for key
```

hazelcast/proxy/vector_collection.py
```diff
@@ -29,29 +29,117 @@
...
+
     def __init__(self, service_name, name, context):
         super(VectorCollection, self).__init__(service_name, name, context)
 
     def blocking(self) -> "BlockingVectorCollection":
+        """Returns a blocking variant of VectorCollection"""
         return BlockingVectorCollection(self)
 
     def get(self, key: Any) -> Future[Optional[Document]]:
+        """Returns the value for the specified key, or ``None`` if this VectorCollection
```
k-jamroz, line 60:
```suggestion
        """Returns Document for the specified key, or ``None`` if this VectorCollection
```

hazelcast/proxy/vector_collection.py
```diff
@@ -91,10 +198,28 @@ def search_near_vector(
...
+        The search is performed using distance metric set when
+        creating the vector index.
+
+        Args:
+            vector: The vector to be used as the reference.
+                It must have the same dimension as specified when creating the vector index.
+            include_value: Return value attached to the Document.
+            include_vectors: Return vectors attached to the Document.
+            limit: Limit the maximum number of Documents returned.
+                If not set, the server default is used.
```
k-jamroz, line 215:
```suggestion
            limit: Limit the maximum number of Documents returned.
```

hazelcast/vector.py
```diff
@@ -24,6 +35,13 @@ class Vector:
 
 
 class Document:
+    """Document represents a value and associated vectors.
+
+    Args:
+        value: The value to associate with this Document.
```
k-jamroz, line 41:
```suggestion
        value: The value associated with this Document.
```

hazelcast/proxy/vector_collection.py
```diff
@@ -29,29 +29,117 @@
 
 
 class VectorCollection(Proxy["BlockingVectorCollection"]):
+    """VectorCollection contains documents with vectors.
+
+    Concurrent, distributed, observable and searchable vector collection.
+    The vector collection can work both async(non-blocking) or sync(blocking).
+    Blocking calls return the value of the call and block the execution until return value is calculated.
+    However, async calls return ``Future`` and do not block execution.
```
srknzl, line 37:
```suggestion
    However, async calls return a ``Future`` and do not block execution.
```

hazelcast/proxy/vector_collection.py
```diff
@@ -29,29 +29,117 @@
...
+    Result of the ``Future`` can be used whenever ready.
+    A ``Future``'s result can be obtained with blocking the execution by calling ``future.result()``.
```
srknzl, line 39: It can also have a callback function to handle it full async, so I believe we should give a link to the Future related documentation from here and that's enough.

"However, async calls return ``Future`` and do not block execution." and a link should be enough

hazelcast/proxy/vector_collection.py
```diff
@@ -80,6 +168,25 @@ def put_all(self, map: Dict[Any, Document]) -> Future[None]:
...
+        already associated.
+
+        Warning:
+            This method returns a clone of the previous Document, not the original
+            (identically equal) Document previously put into the VectorCollection.
+
+        Warning:
+            This method uses ``__hash__`` and ``__eq__`` methods of binary form
+            of the key, not the actual implementations of ``__hash__`` and
+            ``__eq__`` defined in key's class.
```
srknzl, line 181: What does this mean? Isn't this true for all objects for all APIs of the Hazelcast client? We don't specify this for other APIs though. Can it be confusing here as extra?

hazelcast/proxy/vector_collection.py
```diff
@@ -80,6 +168,25 @@ def put_all(self, map: Dict[Any, Document]) -> Future[None]:
...
+
+        Args:
+            key: Key of the entry.
+            document: Document of the entry.
+
+        Returns:
```
srknzl, line 187: if it's absent, how it will return the old Document?

hazelcast/proxy/vector_collection.py
```diff
@@ -91,10 +198,28 @@ def search_near_vector(
...
+        """Returns the Documents closest to the given vector.
+
+        The search is performed using distance metric set when
+        creating the vector index.
+
+        Args:
+            vector: The vector to be used as the reference.
+                It must have the same dimension as specified when creating the vector index.
+            include_value: Return value attached to the Document.
+            include_vectors: Return vectors attached to the Document.
```
srknzl, line 213:
```suggestion
            include_value: Whether to return value attached to the Document.
            include_vectors: Whether to return vectors attached to the Document.
```
I think better, not strong

hazelcast/proxy/vector_collection.py
srknzl, line 215: Why not document the default is 10 here?

hazelcast/vector.py
srknzl: @yuce Do we have python documentation for Type enum(I guess it's a Enum) members? We should have imo

hazelcast/vector.py
```diff
@@ -33,6 +51,11 @@ def __init__(self, value: Any, vectors: Union[Vector, List[Vector]]) -> None:
 
     @property
     def vector(self) -> Optional[Vector]:
+        """Returns the vector associated with this Document
```
srknzl, line 54: I wonder how to access all vectors in the List if user requires that

hazelcast/vector.py
```diff
@@ -68,20 +102,36 @@ class Metric(enum.IntEnum):
...
+    ef_construction: int = 100
+    use_deduplication: bool = True
 
 
 VectorIndexConfig = IndexConfig
 
 
 @dataclasses.dataclass
 class VectorSearchOptions:
+    """VectorSearchOptions contains search configuration options.
```
srknzl, line 131: I think we can make this an inner class in VectorCollection. Not sure if it appears in the python API reference if we do that. It does not have to appear, so we can make this internal if possible

hazelcast/proxy/vector_collection.py
yuce: That's directly from the `IMap` docs, so I'll keep that for consistency.

hazelcast/proxy/vector_collection.py
yuce: That's straight from the `IMap` docs.

hazelcast/proxy/vector_collection.py
yuce: That's straight from the `IMap` docs.

hazelcast/proxy/vector_collection.py
yuce: Took it straight from `Map.put_if_absent`.

yuce: `put_if_absent` checks for equality of the serialized binary object on the server side. Not all methods do that.

hazelcast/proxy/vector_collection.py
yuce: I think both versions are fine.

hazelcast/vector.py
yuce: It's `document.vectors`.

hazelcast/vector.py
```diff
@@ -60,6 +93,7 @@ class SearchResult:
 
 
 class Metric(enum.IntEnum):
+    """Metric is the similarity metric to use for indexing a vector index."""
 
     EUCLIDEAN = 0
     COSINE = 1
```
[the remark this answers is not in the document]
yuce: I think these names are pretty self-explanatory. We can have more in-depth docs, like when would you use Cosine vs Euclidean, but it's probably should be done in the main docs.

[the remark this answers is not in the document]
k-jamroz: these are well known metrics

hazelcast/vector.py
yuce: Why make it an inner class of VectorCollection?

hazelcast/vector.py
yuce: Added some docs at: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/proxy/vector_collection.py
yuce: I've added `optional operation` to the `vc.remove` method for consistency.

hazelcast/proxy/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/vector.py
```diff
@@ -24,6 +35,13 @@ class Vector:
 
 
 class Document:
+    """Document represents a value and associated vectors.
+
+    Args:
+        value: The value to associate with this Document.
+        vectors: Either one ``Vector`` instance of a list of ``Vector`` instances to associate with this Document.
```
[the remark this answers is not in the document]
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/vector.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/proxy/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/proxy/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/proxy/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/proxy/vector_collection.py
yuce: Added the default `10` at:
https://github.com/hazelcast/hazelcast-python-
client/pull/678/commits/75ff67f5ca7811e094e2a3603b04e59f51eeceab

hazelcast/proxy/vector_collection.py
yuce: https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/afce5c2467ca49176d0c1d9b199e65669d2d7268

hazelcast/proxy/vector_collection.py
k-jamroz: shouldn't we DRY? default `limit` is clearly visible as default value of the method parameter

hazelcast/proxy/vector_collection.py
k-jamroz: for consistency with what? Vector colleciton is not IMap and "optional operation" does not mean anything in context of vector collection. We should remove it because it is only misleading and confusing in case of vector collection.

hazelcast/proxy/vector_collection.py
k-jamroz: IMO we do not need typo-level consistency, but we can keep it as is

k-jamroz (changes requested):
Remove meaningless "optional operation" from vector collection docs

hazelcast/proxy/vector_collection.py
yuce: `VectorCollection` is closest to the `IMap. As we've discussed, "optional operation" means the operation becomes no-op if the mapping does not exist. The user wouldn't necessarily know whether VectorCollection uses a map store or not.

Feel free to send another PR to explain this better after this PR is merged.

k-jamroz: > "optional operation" means the operation becomes no-op if the mapping does not exist

this is completely not obvious meaning for me, do you think that our users will understand it that way? @shultseva @vbekiaris how do understand that?

also:

1. there are more operations that do not change state of the system if the collection does not exist, eg. `get`, `search`, `size`, `clear`
2. for all those cases it does not matter if the collection exists or not (strictly speaking if `DistributedObject` exists) but if the collection (or partition for `get`) is empty or not (collection that does not exist is empty by definition). And the result is logically obvious: if the collection is empty then removing from it does not change it
3. we want to cut ties between vector collection and IMap, there is no value is artificially making them similar

so this is a useless, confusing information.

hazelcast/vector.py
```diff
@@ -60,6 +96,7 @@ class SearchResult:
 
 
 class Metric(enum.IntEnum):
+    """Metric is the similarity metric to use for indexing a vector index."""
```
EugeneAbramchuk, line 99: This definition is somehow very convoluted. I am sure it can be improved but I am unsure how.

I propose to start with 
`Metric is the measure of similarity which is used for vector indexing to enable searching.`

The docs team can likely come up with something better

hazelcast/vector.py
```diff
@@ -68,20 +105,36 @@ class Metric(enum.IntEnum):
...
+
+    Args:
+        name: Name of the vector index.
+        metric: The metric to be used for the index configuration.
+        dimension: The dimension of vectors to be used in the vector index.
+            All vectors that use this index configuration must have
+            the same dimension.
+        max_degree: The maximum number of connections allowed per node.
+        ef_construction: The size of the dynamic list for search.
+        use_deduplication: Enable deduplication in the index.
```
EugeneAbramchuk, line 118: I would rather add some explanation what deduplication means, but there is no mention of it anywhere in the docs elsewhere... (neither in the java docs)

k-jamroz: it is described in https://deploy-preview-1124--hardcore-allen-f5257d.netlify.app/hazelcast/5.5-snapshot/data-structures/vector-collections#configuration (not yet merged)

hazelcast/vector.py
yuce: Not sure your suggestion is more understandable than the previous one :)

hazelcast/proxy/vector_collection.py
```diff
@@ -29,29 +29,118 @@
...
         check_not_none(key, "key can't be None")
         check_not_none(document, "document can't be None")
         check_not_none(document.value, "document value can't be None")
         return self._set_internal(key, document)
 
     def put(self, key: Any, document: Document) -> Future[Optional[Document]]:
+        """Associates the specified Document with the specified key in this VectorCollection.
+
+        If the VectorCollection previously contained a mapping for the key, the old Document is
+        replaced by the specified Document.
```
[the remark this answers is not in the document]
yuce: I changed it as follows:
https://github.com/hazelcast/hazelcast-python-client/pull/678/commits/6fb9acf06987c07ea7d93941221a316818aa8514

hazelcast/proxy/vector_collection.py
yuce: The defaults are explicitly mentioned in the documentation in some other data structure methods, like `Map.add_index`, so I think it's OK to keep it.

hazelcast/vector.py
EugeneAbramchuk: I agree my suggestion is not an improvement, I was waiting for the docs team to comment. We can pick this up later and merge as is for now

EugeneAbramchuk (approved):

srknzl (approved):

k-jamroz (commented):
my [comment](https://github.com/hazelcast/hazelcast-python-client/pull/678/files#r1647615569) about "optional operation" was not addressed, but if others are happy with it, I will not block merging this PR

k-jamroz (approved):

srknzl:
> my [comment](https://github.com/hazelcast/hazelcast-python-client/pull/678/files#r1647615569) about "optional operation" was not addressed, but if others are happy with it, I will not block merging this PR

I also think we should remove it if it's not obvious and creates confusion. However, maybe we should do it in general regarding all python docstrings in the repo

yuce:
It seems tests jars for snapshots are not produced anymore. Because of that the test runner fails.
I've raised: https://hazelcast.atlassian.net/browse/HZAI-101 for that.

PR #679, opened 2024-06-14
Update vector collection API - search, optimize, clear, size

k-jamroz:
Implement client side changes related to recent client protocol changes and some cleanup:
- move vector to top-level search parameter
- add search hints (minimal API with `Dict[str, str]`, can be improved later to be more python-like)
- add `optimize` and `clear`
- add `size`
- fix automatically generated codecs (https://github.com/hazelcast/hazelcast-client-protocol/pull/528)
- additional vector JARs are no longer needed
- add `Document.__repr__` for easier printing
- removed default search limit, `-1` was invalid value
- use V7 license key for tests

### DIFF
--- a/hazelcast/proxy/vector_collection.py
+++ b/hazelcast/proxy/vector_collection.py
@@ -90,7 +90,7 @@ class VectorCollection(Proxy["BlockingVectorCollection"]):
         *,
         include_value: bool = False,
         include_vectors: bool = False,
-        limit: int = -1
+        limit: int
     ) -> Future[List[SearchResult]]:
         check_not_none(vector, "vector can't be None")
         return self._search_near_vector_internal(
@@ -152,7 +152,7 @@ class VectorCollection(Proxy["BlockingVectorCollection"]):
         *,
         include_value: bool = False,
         include_vectors: bool = False,
-        limit: int = -1
+        limit: int
     ) -> Future[List[SearchResult]]:
         def handler(message):
             results: List[
@@ -259,7 +259,7 @@ class BlockingVectorCollection:
         *,
         include_value: bool = False,
         include_vectors: bool = False,
-        limit: Optional[int] = None
+        limit: int
     ) -> List[SearchResult]:
         future = self._wrapped.search_near_vector(
             vector,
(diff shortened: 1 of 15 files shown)


k-jamroz:
In order for vector collection tests to pass we need V7 license

hazelcast/proxy/vector_collection.py
```diff
@@ -106,6 +111,18 @@ def delete(self, key: Any) -> Future[None]:
...
 
+    def optimize(self, index_name: str = None) -> Future[None]:
+        request = vector_collection_optimize_codec.encode_request(self.name, index_name)
+        return self._invoke(request)
+
+    def clear(self) -> Future[None]:
+        request = vector_collection_clear_codec.encode_request(self.name)
+        return self._invoke(request)
+
+    def size(self) -> Future[int]:
```
yuce, line 122: Could you add API docs to these methods?

hazelcast/proxy/vector_collection.py
```diff
@@ -142,7 +159,8 @@ def _search_near_vector_internal(
         *,
         include_value: bool = False,
         include_vectors: bool = False,
-        limit: int = -1
+        limit: int,
+        hints: Dict[str, str] = None
```
yuce, line 163: I think we shouldn't add hints to the Python API.
But let's discuss.

hazelcast/vector.py
```diff
@@ -40,6 +40,9 @@ def vector(self) -> Optional[Vector]:
     def __copy__(self):
         return Document(self.value, self.vectors)
 
+    def __str__(self):
+        return self.value + " " + str(self.vectors)
```
yuce, line 44:
```suggestion
    def __repr__(self):
        return f"Vector<value={self.value}, vectors=self.vectors}>"
```

yuce (changes requested):

hazelcast/proxy/vector_collection.py
k-jamroz: https://github.com/hazelcast/hazelcast-python-client/pull/679/commits/a76fc95fb840fe6b5338097d0fefbabaccf514b5

hazelcast/proxy/vector_collection.py
k-jamroz: as discussed today: hints will be available in python API, `partitionLimit` hint vs explicit parameter to discuss (outside this PR), better validation/conversion of hints (eg. converting number values to strings) can be implemented also outside this PR

yuce (dismissed):
Looks good.

yuce (approved):

PR #684, opened 2024-07-09
Use the private snapshot repo

yuce:
* Updates the artifact downloader to use snapshots from the private repo.
* Mark more CP tests as enterprise

The tests fail since the artifacts cannot be downloaded.
Once the PR is merged, H workflow updates will be in-effect, and the artifacts will be downloaded fine.

### DIFF
--- a/.github/workflows/coverage_runner.yml
+++ b/.github/workflows/coverage_runner.yml
@@ -75,9 +75,9 @@ jobs:
           
       - name: Run tests
         env:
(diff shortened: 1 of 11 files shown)


emreyigit (approved):
Is `mvn` installed on the action? I couldn't see it. OTT, looks great.

yuce:
> Is `mvn` installed on the action? I couldn't see it. OTT, looks great.

The GH Java action installs Maven.

PR #711, opened 2024-11-14
Add tests for Vector Collection backup config

olukas:
Add tests for setting Vector Collection `backup_count` and `async_backup_count` configuration.

### DIFF
--- a/tests/integration/backward_compatible/proxy/vector_collection_test.py
+++ b/tests/integration/backward_compatible/proxy/vector_collection_test.py
@@ -181,7 +181,21 @@ class VectorCollectionTest(SingleMemberTestCase):
             name, [IndexConfig("vector", Metric.COSINE, 3)], backup_count=2, async_backup_count=2
         )
 
-    def test_backupCount(self):
+    def test_backupCount_max_value_pass(self):
+        skip_if_client_version_older_than(self, "6.0")
+        name = random_string()
+        self.client.create_vector_collection_config(
+            name, [IndexConfig("vector", Metric.COSINE, 3)], backup_count=6
+        )
+
+    def test_backupCount_min_value_pass(self):
+        skip_if_client_version_older_than(self, "6.0")
+        name = random_string()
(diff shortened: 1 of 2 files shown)


tests/integration/backward_compatible/proxy/vector_collection_test.py
```diff
@@ -181,7 +181,21 @@ def test_backupCount_valid_values_pass(self):
             name, [IndexConfig("vector", Metric.COSINE, 3)], backup_count=2, async_backup_count=2
         )
 
-    def test_backupCount(self):
+    def test_backupCount_max_value_pass(self):
+        skip_if_client_version_older_than(self, "6.0")
+        name = random_string()
+        self.client.create_vector_collection_config(
+            name, [IndexConfig("vector", Metric.COSINE, 3)], backup_count=6
+        )
```
yuce, line 189: Would it make sense to check whether the VC was created, maybe something like:
```python
client.get_vector_collection(vc_name).blocking()
```
That would raise an exception if the vector collection doesn't exist.
Same comment for tests that creates a VC.

tests/integration/backward_compatible/proxy/vector_collection_test.py
```diff
@@ -181,7 +181,21 @@ def test_backupCount_valid_values_pass(self):
             name, [IndexConfig("vector", Metric.COSINE, 3)], backup_count=2, async_backup_count=2
         )
 
-    def test_backupCount(self):
+    def test_backupCount_max_value_pass(self):
```
yuce, line 184: Would it be possible to use lower snake case exclusively for function names, like `test_backup_count_max_value_pass`?
That's the expected Python style.
I've missed that in previous that tests that involve the backup count.
I'll send a PR to fix them.

yuce (changes requested):
Just a few ideas...

yuce (dismissed):
Looks good!

yuce (approved):

srknzl (approved):

PR #719, opened 2025-02-11
Centralise Python version specification in GitHub Actions

JackPGreen:
We duplicate the specification of Python versions in each of the actions - instead we can centralise it to a single file that all actions can use.

Examples showing the version resolution (not necessarily the actions themselves!) working correctly:
- https://github.com/hazelcast/hazelcast-python-client/actions/runs/13271945670
- https://github.com/hazelcast/hazelcast-python-client/actions/runs/13271945671

Additional changes:
- upgraded setup-python action to latest
- removed [PyPy exclusion](https://github.com/hazelcast/hazelcast-python-client/pull/625) because [PyPy was removed from execution](https://github.com/hazelcast/hazelcast-python-client/pull/697).


.github/workflows/get-python-versions.yml
```diff
@@ -0,0 +1,28 @@
...
+  get-python-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      python-versions: ${{ steps.extract-versions.outputs.python-versions }}
+      earliest-python-version: ${{ steps.extract-versions.outputs.earliest-python-version }}
+      latest-python-version: ${{ steps.extract-versions.outputs.latest-python-version }}
+    steps:
+      - id: extract-versions
+        run: |
+          python_versions='[ "3.8", "3.9", "3.10", "3.11", "3.12" ]'
```
yuce, line 23: Could you also add `3.13` there?

yuce (changes requested):
Just a single idea...

.github/workflows/get-python-versions.yml
JackPGreen: I’ve no idea adding 3.13 - but could I raise a follow up PR to do that separately?

yuce: Sure.

Issue #746, opened 2025-10-13
RFE client.py: get_distributed_objects should skip cacheService


space88man:
RFE: `client.py:get_distributed_objects` will fail if the cluster is also used by Hazelcast/JCache as it doesn't support cacheService. Tweak `get_distributed_objects` to skip cacheService instead of throwing an exception.

```python
    def get_distributed_objects(self) -> Future[typing.List[Proxy]]:
        """Returns all distributed objects such as; queue, map, set, list,
        topic, lock, multimap.

        Also, as a side effect, it clears the local instances of the destroyed
        proxies.

        Returns:
            List of instances created by Hazelcast.
        """
        request = client_get_distributed_objects_codec.encode_request()
        invocation = Invocation(request, response_handler=lambda m: m)
        self._invocation_service.invoke(invocation)

        local_distributed_object_infos = {
            DistributedObjectInfo(dist_obj.service_name, dist_obj.name)
            for dist_obj in self._proxy_manager.get_distributed_objects()
        }

        response = client_get_distributed_objects_codec.decode_response(invocation.future.result())
        for dist_obj_info in response:
            local_distributed_object_infos.discard(dist_obj_info)

            # we need to skip JCache service objects
            # otherwise will throw an exception here
            if dist_obj_info.service_name == "hz:impl:cacheService":
                continue
            self._proxy_manager.get_or_create(
                dist_obj_info.service_name, dist_obj_info.name, create_on_remote=False
            )

        for dist_obj_info in local_distributed_object_infos:
            self._proxy_manager.destroy_proxy(
                dist_obj_info.service_name, dist_obj_info.name, destroy_on_remote=False
            )

        return self._proxy_manager.get_distributed_objects()
```

PR #747, opened 2025-10-14
get_distributed_objects: skip unsupported cacheService

space88man:
This client does not support cacheService so the iteration in `get_distributed_objects` will throw an exception if the cluster is also used by Java+JCache clients (e.g. Spring Boot with Hazelcast as cache provider).

This fix just skips `hz:impl:cacheService` objects.

Addresses #746


yuce:
Thanks!
We'll need to fix the code style in `hazelcast/client.py`.
Could you `pip install -r requirements-dev.txt` then run `black --config black.toml .` in the project root?
That should fix the issue.

space88man:
black formatting is fixed  - PTAL

space88man:
> ... then run `black --config black.toml .` in the project root? That should fix the issue.

Done - PTAL

PR #780, opened 2026-02-19
[HZ-5398] Asyncio Client Fixes (2)

yuce:
Some small and big fixes and improvements to mostly the asyncio client, but some asyncore as well:

* Check the existence of event handlers in *Map proxies.
* Cluster get_member must use the correct variable
* Compact schema send retry, blocking cloud discovery refresh
* Create tasks for conn.close_connection calls.
* Fix ConnectionManager start_connect_all_members task
* Fixed the deadlock in asyncio listener
* Fixed the race condition in ProxyManager.get_or_create
* Fixed the race condition in _handle_successful auth
* Fixed wrong method retries in VC
* Remove the write loop in reactor
* VC put_if_absent must copy the passed document before updating it.

### DIFF
--- a/hazelcast/internal/asyncio_cluster.py
+++ b/hazelcast/internal/asyncio_cluster.py
@@ -262,7 +262,7 @@ class _InternalClusterService:
                 if handler:
                     try:
                         handler(dead_member)
-                    except:
+                    except Exception:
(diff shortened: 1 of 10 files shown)


emreyigit (approved):
LGTM.

PR #781, opened 2026-02-20
[HZ-5399] Optimize Serialization

yuce:
This PR optimizes array serialization by preallocating the necessary space and  reading all array items all at once, instead of reading them one by one.


ihsandemir (commented):
Did you performa any performance analysis to prove that this refactoring actually improves? do you have such data to support the change?

ihsandemir (approved):

PR #790, opened 2026-03-23
[HZ-5405] [HZ-5419] Executor for Asyncio

yuce:
Straightforward port of Executor proxy, and its tests to asyncio

### DIFF
--- a/hazelcast/internal/asyncio_proxy/base.py
+++ b/hazelcast/internal/asyncio_proxy/base.py
@@ -67,6 +67,12 @@ class Proxy(typing.Generic[BlockingProxyType], abc.ABC):
         self._invocation_service.invoke(invocation)
         return invocation.future
 
+    async def _ainvoke_on_target(
+        self, request, uuid, response_handler=_no_op_response_handler
+    ) -> typing.Any:
+        fut = self._invoke_on_target(request, uuid, response_handler)
+        return await fut
+
     async def _invoke_on_key(
         self, request, key_data, response_handler=_no_op_response_handler
     ) -> typing.Any:
--- a/hazelcast/internal/asyncio_proxy/executor.py
+++ b/hazelcast/internal/asyncio_proxy/executor.py
@@ -85,11 +85,10 @@ class Executor(Proxy):
         except SchemaNotReplicatedError as e:
             return await self._send_schema_and_retry(e, self.execute_on_members, members, task)
 
-        uuid = uuid4()
         tasks = []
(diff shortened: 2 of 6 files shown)


hazelcast/internal/asyncio_proxy/base.py
```diff
@@ -67,6 +67,12 @@ def _invoke_on_target(
         self._invocation_service.invoke(invocation)
         return invocation.future
 
+    async def _ainvoke_on_target(
```
ihsandemir, line 70: method only called from one place, `Executor._execute_on_member`. why do you introduce yet another utility method like this instead of awaiting at the called method?

hazelcast/internal/asyncio_proxy/executor.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Returns:
+            The list of results of the tasks on each member.
+        """
+        try:
+            task_data = self._to_data(task)
+        except SchemaNotReplicatedError as e:
+            return await self._send_schema_and_retry(e, self.execute_on_members, members, task)
+
+        uuid = uuid4()
```
ihsandemir, line 88: The uuid is unique for each call to each member at Java client(see https://github.com/hazelcast/hazelcast-mono/blob/d2d3c923d45982e80047954849a560dc85762821/hazelcast/hazelcast/src/main/java/com/hazelcast/client/impl/proxy/ClientExecutorServiceProxy.java#L511). Can you test and fix this same as Java?

ihsandemir (changes requested):

hazelcast/internal/asyncio_proxy/base.py
yuce: It's similar to `_ainvoke_on_partition`: https://github.com/yuce/hazelcast-python-client/blob/4be241441bf1cc9e54a8bc03e4927986f32ef576/hazelcast/internal/asyncio_proxy/base.py#L94 

The `_invoke_on_partition` method returns a `Future`, so if you call it in an `async` function, there's no linter check to make sure you called it with `await`. I introduced `_ainvoke_target`, so it is clear that it's an async function, and linter shows a warning if it wasn't called with `await`.

hazelcast/internal/asyncio_proxy/executor.py
yuce: Fixed both asncore and asyncio code at: https://github.com/hazelcast/hazelcast-python-client/pull/790/commits/c7e6979c7a9f94017e73b3a66276763d6c4fe236

ihsandemir (approved):

PR #797, opened 2026-04-02
[HZ-5407] Topic for Asyncio

yuce:
Straightforward port of Topic proxy to Asyncio Client.

### DIFF
--- a/hazelcast/internal/asyncio_proxy/topic.py
+++ b/hazelcast/internal/asyncio_proxy/topic.py
@@ -1,19 +1,19 @@
 import typing
 
-from hazelcast.future import Future
 from hazelcast.protocol.codec import (
     topic_add_message_listener_codec,
     topic_publish_codec,
     topic_publish_all_codec,
     topic_remove_message_listener_codec,
 )
-from hazelcast.proxy.base import PartitionSpecificProxy, TopicMessage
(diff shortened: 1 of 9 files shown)


hazelcast/internal/asyncio_proxy/topic.py
```diff
@@ -0,0 +1,116 @@
...
+    Example:
+        >>> my_topic = await client.get_topic("my_topic")
+        >>> await my_topic.publish("hello")
+
+    Warning:
+        Asyncio client topic proxy is not thread-safe, do not access it from other threads.
+    """
+
+    async def add_listener(
+        self, on_message: typing.Callable[[TopicMessage[MessageType]], None] = None
```
emreyigit, line 38: I know this is from sync version but a check against to null handler would be nice on both version. 

Also, how is `on_message` going to be invoked? Is it sync or async? If it is async, documenting it would be nice, IMO. 

Rest is LGTM.

emreyigit (approved):

hazelcast/internal/asyncio_proxy/topic.py
yuce: Currently the callback must be non-blocking.
I've added the check and doc at: https://github.com/hazelcast/hazelcast-python-client/pull/797/commits/570d825cbb4b7124f5df8206d3f46ba49007995b

PR #818, opened 2026-06-08
[HZ-5434] FencedLock for Asyncio

yuce:
Ported `FencedLock` proxy to asyncio.

The asyncore variant uses actual thread IDs to disable re-entrancy.
That's not useful with the asyncio variant, so currently the thread ID `0` is used when necessary.
I've omitted docs that refer to threads, and tests that use threads.
I'll push another PR to add those after lock context implementation.


emreyigit:
How are you going to protect lock ownership without thread id? There will be concurrent runs, right?

yuce:
> How are you going to protect lock ownership without thread id? There will be concurrent runs, right?

The lock is reentrant for lock attempts with the same client, similar to the Node.js client, so data races will happen with operations that uses the same client.
I'll push another PR that uses a pseudo-task ID to be used instead of the thread ID, which will enable us to distinguish between tasks.

yuce:
The SSL tests with Python 3.13 + Windows fail, but that's unrelated to this PR.
Both asyncore and asyncio SSL tests fail.

emreyigit (approved):

